PR #67, opened 2020-08-28
Feature/integrate offline queue

dron22:
This PR integrates the offline queue into send heartbeats command. The following flags are added:

- `disable-offline`
- `disableoffline` (deprecated)
- `sync-offline-activity`

Alternatively `offline` boolean can be defined in `settings` section of `.wakatime.cfg` file. Default setting for number of heartbeats synced from offline queue is `100`, like in the python version.

### DIFF
--- a/cmd/legacy/heartbeat/params.go
+++ b/cmd/legacy/heartbeat/params.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -27,19 +28,21 @@ var (
 
 // Params contains heartbeat command parameters.
 type Params struct {
-	APIKey     string
-	APIUrl     string
-	Category   heartbeat.Category
-	Entity     string
-	EntityType heartbeat.EntityType
-	Hostname   string
-	IsWrite    *bool
-	Plugin     string
-	Time       float64
-	Timeout    time.Duration
-	Filter     FilterParams
-	Network    NetworkParams
-	Sanitize   SanitizeParams
+	APIKey          string
+	APIUrl          string
+	Category        heartbeat.Category
+	Entity          string
+	EntityType      heartbeat.EntityType
+	Hostname        string
+	IsWrite         *bool
+	OfflineDisabled bool
+	OfflineSyncMax  int
+	Plugin          string
+	Time            float64
+	Timeout         time.Duration
+	Filter          FilterParams
+	Network         NetworkParams
+	Sanitize        SanitizeParams
 }
 
 // FilterParams contains heartbeat filtering related command parameters.
@@ -93,7 +96,7 @@ func LoadParams(v *viper.Viper) (Params, error) {
 	}
 
 	entity, ok := vipertools.FirstNonEmptyString(v, "entity", "file")
-	if !ok {
+	if !ok && !v.IsSet("sync-offline-activity") {
 		return Params{}, errors.New("failed to retrieve entity")
 	}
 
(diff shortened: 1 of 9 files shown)


cmd/root.go
```diff
@@ -17,6 +17,9 @@ const (
 	defaultConfigSection = "settings"
 	// defaultTimeoutSecs is the default timeout used for requests to the wakatime api
 	defaultTimeoutSecs = 60
+	// defaultOfflineSync is the default maximum number of heartbeats from the
+	// offline queue, which will be synced upon sending heartbeats to the API.
```
gandarez, line 21: Ponctuation at the end is missing in the other comments. Should it be removed from this line?

pkg/offline/offline.go
```diff
@@ -5,16 +5,40 @@ import (
...
 
+// QueueFilepath returns the path to the offline queue db file.
+func QueueFilepath() (string, error) {
+	dir := os.Getenv("WAKATIME_HOME")
+
+	var err error
+	if dir == "" {
+		dir, err = os.UserHomeDir()
+		if err != nil {
+			return "", fmt.Errorf("failed to retrieve user's home dir: %s", err)
```
gandarez, line 30: just to follow the other messages like that it should be `failed getting user's home directory: %s`

pkg/offline/offline.go
```diff
@@ -5,16 +5,40 @@ import (
...
+		}
+	}
+
+	expanded, err := homedir.Expand(dir)
+	if err != nil {
+		return "", fmt.Errorf("failed to expand offline queue folder path: %s", err)
```
gandarez, line 36: maybe `failed expanding offline queue folder path`?

cmd/root.go
dron22: Actually the linter should enforce punctuation at the end of comments. I suspect that the linter does not check comments on non-exported (lower-case) names. Nevertheless, let's keep punctuation and also add it to the other comments.

pkg/offline/offline.go
dron22: Actually the other messages in the package are all in this format (`failed to ...`). Can you point me to where we use the other format?

cmd/legacy/heartbeat/params.go
```diff
@@ -123,6 +126,30 @@ func LoadParams(v *viper.Viper) (Params, error) {
 		isWrite = heartbeat.Bool(b)
 	}
 
+	offlineDisabled := vipertools.FirstNonEmptyBool(v, "disableoffline", "disable-offline")
```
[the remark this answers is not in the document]
dron22: This is the behavior of the python client. If you look at https://github.com/wakatime/wakatime/blob/master/wakatime/arguments.py#L340, you can see, that `args.offline` contains the `disable-offline` value. But `args.offline_deprecated` will overwrite it.

I agree that what you suggest is actually more appropriate. Should we ask @alanhamlett for an opinion here?

dron22:
@gandarez FYI, I added handling of heartbeat leftovers to the offline queue. This was missing in offline queue implementation and first I planned to push this in a follow up PR, but now I added it here.

cmd/legacy/heartbeat/params.go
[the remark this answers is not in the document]
dron22: @alanhamlett In the python client, if both set, the deprecated `disableoffline` flag overwrites `disable-offline` flag.

What would you say regarding this edge case? Should we keep the behavior from the python version or adjust this?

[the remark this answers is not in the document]
alanhamlett: I think for consistency let's keep the same logic as the Python cli. We'll start breaking with the Python cli in future releases, after making sure everything works the same way first.

dron22:
FYI, the PR contains conflicting files, because filestats PR (#68) was merged recently, which also added flag params. Let's keep the commit history and conversation of the PR and finish the review/approval for the solution first. Then I'll rebase before merging it to master to remove the conflicts.

PR #68, opened 2020-08-28
filestats detection

dron22:
This PR add `filestats` package and integrates it into send heartbeat command.

- This will detect the total number of lines in the file associated with a heartbeat.
- `CursorPosition` and `LineNumbers` are directly taken from flags, just like in the python version.

Flags added:
- `cursorpos`
- `lineno`


gandarez:
Let's only review it after merging https://github.com/wakatime/wakatime-cli/pull/67

dron22:
@gandarez As there is a question to Alan in #67 PR, I rebased this one, so it does not build on offline queue changes anymore. Can be reviewed now.

pkg/filestats/filestats.go
```diff
@@ -0,0 +1,74 @@
...
+		return func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
+			for n, h := range hh {
+				if h.EntityType == heartbeat.FileType {
+					fileInfo, err := os.Stat(h.Entity)
+					if err != nil {
+						jww.ERROR.Printf("failed to retrieve file stats of file %q: %s", h.Entity, err)
+						continue
+					}
+
+					if fileInfo.Size() > maxFileSizeSupported {
```
gandarez, line 32: Isn't there a way to test this validation? I believe uploading a file with more than 2mb to test it will increase the download time for this repository. What do you think?

gandarez (commented):
In general looks good. Just one question regarding the tests.

pkg/filestats/filestats.go
dron22: @gandarez I added a test, which will create a temporary file, which exceeds the size.

gandarez: good!

gandarez (approved):
looks good

PR #69, opened 2020-08-31
add subversion project detection

gandarez:
This PR adds subversion project detection.


pkg/project/testdata/svn/format
```diff
@@ -0,0 +1 @@
+12
```
dron22, line 1: The svn binary depends on all these files?

dron22 (changes requested):
For me the tests fail locally, because of missing svn binary. Why are they even passing on github? Does github has the svn binary installed on the github action container? Or is the test not run at all?

Would be good, if you check this. Regarding binary, we might need to add it to the codebase. How big is this one? If it is in `testdata` folder, it won't be included in the final built binary.

pkg/project/testdata/svn/format
gandarez: Thanks for asking it. I'd never tried removing files from `SVN` but removed some of them and the tests are still passing.

gandarez:
> For me the tests fail locally, because of missing svn binary. Why are they even passing on github? Does github has the svn binary installed on the github action container? Or is the test not run at all?
> 
> Would be good, if you check this. Regarding binary, we might need to add it to the codebase. How big is this one? If it is in `testdata` folder, it won't be included in the final built binary.

We use ubuntu 18.04 and there's svn client installed on it as you can see in this documentation https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md

The `svn` binary is 31kb in size but it's compiled per operation system. Because of that I can't simply copy it to `testdata` folder. What about skipping tests locally when the binary is not found?

dron22:
Skipping test conditionally sounds good to me.

dron22 (approved):

PR #329, opened 2021-05-02
Do not build binaries on alpha releases

### LINKED ISSUE #328: Do not build binaries on alpha releases
To not exhaust our GH storage, binaries should not be build on alpha releases.

dron22:
This PR changes `on_release` workflow, to not run on prereleases. This is achieved by adding a condition to `on_release` workflow to skip it for prereleases.

To continue notifying to slack on alpha releases, a new workflow `on_prerelease` is added.

Closes #328

### DIFF
--- a/.github/workflows/on_prerelease.yaml
+++ b/.github/workflows/on_prerelease.yaml
@@ -1,0 +1,33 @@
+name: Notify about new prerelease
+
+on:
+  release:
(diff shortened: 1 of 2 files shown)


alanhamlett:
We can re-open if needed, but there shouldn't be any limits now that the repo is public.

PR #338, opened 2021-05-03
Add coverage report

gandarez:
This PR replaces the inactive and outdated https://github.com/wakatime/wakatime-cli/pull/294.


.github/workflows/on_push_protected_branches.yml
```diff
@@ -30,6 +30,11 @@ jobs:
       - 
         name: "Linter"
         run: make lint
+      -
+        name: "Send coverage"
+        uses: shogo82148/actions-goveralls@v1
+        with:
+          path-to-profile: coverage.out
```
dron22, line 37: I guess we need to generate the coverage profile, when running tests, right? Compare: https://github.com/mattn/goveralls/blame/master/README.md#L65

README.md
```diff
@@ -1,5 +1,10 @@
 # WakaTime CLI
 
+| branch  | build | coverage |
```
dron22, line 3: The coverage of `develop` branch is mainly for internal purposes and I guess we don't need to have it on top of the README. Let's keep it clean. Badge for master would be enough here I guess.

.github/workflows/on_push_protected_branches.yml
gandarez: exactly and the coverage profile is being generated by make lint, see the changes on MAKEFILE.

README.md
gandarez: As we want to `develop` branch for testing and alpha releases I believe having it at the top would be useful for people who want to make sure build is passing. If we add it why not to also have its coverage reported and exposed?

README.md
```diff
@@ -1,5 +1,10 @@
...
+| ---     | ---   | ---      |
+| develop | ![Build develop](https://img.shields.io/github/workflow/status/wakatime/wakatime-cli/Create%20Release/develop) | [![Coverage Status](https://coveralls.io/repos/github/wakatime/wakatime-cli/badge.svg?branch=develop)](https://coveralls.io/github/wakatime/wakatime-cli?branch=develop)
```
dron22, line 5: `Create Release` workflow actually does not represent the build process.

gandarez: What about having `Create Release` as **Unit Tests Status** and `Build and upload release assets` as **Build Status**?

dron22: Yeah, this sounds like the right way :+1:

dron22 (approved):
As you removed the `develop` branch stats, we don't really need the markdown table anymore, right? We could have the badges without the table on the top, like most repos do it.

Not explicitly mentioning master branch would be alright I guess, as it is safe to assume that the stats represent the latest official release.

Approving already..

Issue #112, opened 2020-11-04
Implement unofficial features to ini parser


gandarez:
The Python's standard library ini parsers supoorts parsing a key with multiple lines as described below. The one we've been using [go-ini](https://github.com/go-ini/ini) does not support it as reported in this [issue](https://github.com/go-ini/ini/issues/270). This feature need to be supported to not break any customer who's already using it.

```ini
[settings]
ignore  =
  COMMIT_EDITMSG$
  PULLREQ_EDITMSG$
  MERGE_MSG$
  TAG_EDITMSG$
[git]
submodules_disabled = 
  .*secret.*
  fix.*
```

gandarez:
Closed by https://github.com/spf13/viper/pull/1156

PR #341, opened 2021-05-03
Add test for multine submodules

gandarez:
This PR removes test skipping from submodules.

ref https://github.com/wakatime/wakatime-cli/issues/112


dron22 (approved):
:+1:

PR #353, opened 2021-05-05
Fix alternate-language param info

dron22:
This PR fixed the info for `--alternate-language` parameter.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -39,8 +39,8 @@ func NewRootCMD() *cobra.Command {
 
 func setFlags(cmd *cobra.Command, v *viper.Viper) {
 	flags := cmd.Flags()
-	flags.String("alternate-language", "", "(deprecated)")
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -39,7 +39,7 @@ func NewRootCMD() *cobra.Command {
 
 func setFlags(cmd *cobra.Command, v *viper.Viper) {
 	flags := cmd.Flags()
-	flags.String("alternate-language", "", "(deprecated)")
+	flags.String("alternate-language", "", "Optional alternate language name. Auto-discovered language takes priority.")
```
gandarez, line 42: What about?
```
Optional language name. Auto-detected language takes priority.
```

dron22: Yeah this sounds better. Will update along with `alternate-project` info, to keep it consistent :+1:

gandarez (approved):

Issue #352, opened 2021-05-05
Don't print help text of deprecated flags


gandarez:
Deprecated flags might not be printed to the output and marked as deprecated.

PR #354, opened 2021-05-05
Mark deprecated flags

gandarez:
This PR marks deprecated flags to be hidden when executing the cli.

The deprecated `alternate-language` flag will be tackled by @dron22 in another PR.

Closes https://github.com/wakatime/wakatime-cli/issues/352

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -88,7 +88,11 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
 		"When set, any activity where the project cannot be detected will be ignored.",
 	)
 	flags.Bool("extra-heartbeats", false, "Reads extra heartbeats from STDIN as a JSON array until EOF.")
-	flags.String("file", "", "(deprecated) Absolute path to file for the heartbeat.")
+	flags.String(
+		"file",
+		"",
+		"(deprecated) Absolute path to file for the heartbeat."+
+			" Can also be a url, domain or app when --entity-type is not file.")
 	flags.String("hide-branch-names", "", "Obfuscate branch names. Will not send revision control branch names to api.")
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -65,7 +65,7 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
 	)
 	flags.Int("cursorpos", 0, "Optional cursor position in the current file.")
 	flags.Bool("disable-offline", false, "Disables offline time logging instead of queuing logged time.")
-	flags.Bool("disableoffline", false, "(deprecated) Disables offline time logging instead of queuing logged time.")
+	flags.Bool("disableoffline", false, "")
```
dron22, line 68: The info will be suppressed and replaced by `please use --disable-offline instead.`, as defined via `flags.MarkDeprecated()`, right?

Then we should not change the old info. As we still have the logic to handle these parameters, it is good to keep it for documentation purposes. Also the `(deprecated)` info is a good indicator, to spot deprecated params within the list.

dron22 (changes requested):
Nice one :+1: Yeah, let's not tempt users to use the deprecated parameters.

Let's just keep the old info strings untouched.

cmd/root.go
gandarez: Because deprecated flags are hidden the old info might be supressed (removed) as well, it becomes useless since it won't be printed anywhere. The text `please use ....` will be in place when those flags are used.

dron22: That's what I mean. It won't be visible to the user, but helps us as the maintainers, keep track of what these params are doing.

gandarez: got it

dron22 (approved):
LGTM :+1:

Issue #359, opened 2021-05-06
High CPU usage


gandarez:
High CPU usage reported by customers running on Windows machine.

PR #361, opened 2021-05-06
Fix infinite loop for Windows

gandarez:
This PR fixes an infinite loop that may occur for some Windows users when detecting project.

Packages `git` and `mercurial` recursively looks for its configuration file and in windows machines the last path is a `C:\` (or any other drive letter) instead of only a dash. 

Closes https://github.com/wakatime/wakatime-cli/issues/359

### DIFF
--- a/pkg/project/git.go
+++ b/pkg/project/git.go
@@ -154,7 +154,7 @@ func findGitConfigFile(fp string, directory string, match string) (string, bool)
 	}
 
 	dir := filepath.Clean(filepath.Join(fp, ".."))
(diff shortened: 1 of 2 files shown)


dron22 (approved):
Works for me in windows vm :+1:

gandarez:
Also find the same bug for `git` detection.

dron22 (commented):
You describe what this PR does. Could you add some more info to the description on why and how it fixes the issue? Maybe with some example?

dron22 (approved):

PR #362, opened 2021-05-06
Log improvements

dron22:
This PR improves logging. These changes proved to be helpful on debugging the windows recursion problem (#361):

- Allow logging to stdout via `--log-to-stdout` parameter
- Indicate execution of each heartbeat processing pipeline step via debug log
- Allow non-standard binary name on building via make

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := build-all
 
 # globals
-BINARY_NAME=wakatime-cli
+BINARY_NAME?=wakatime-cli
 COMMIT?=$(shell git rev-parse --short HEAD)
 DATE?=$(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
 REPO=github.com/wakatime/wakatime-cli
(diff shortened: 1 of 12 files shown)


gandarez:
Thinking here to make logging more flexible we might want to log to file and to terminal at the same time. When set instead of bypassing logging to file it sets multiple writers. Or even better would be if we define an input to accept an slice of integer/string to determine the outputs. What do you think?

```
mw := io.MultiWriter(os.Stdout, logFile)
logrus.SetOutput(mw)
```

dron22:
@gandarez As alan replied in slack, logging to file and stdout by default wouldn't be an option. My general motivation is to keep it simple. Logging to stdout is helpful for debugging and prevents that logs of a test binary and logs of the main binary mix up.

What would be a scenario, where we want to log to file and stdout at the same time?

gandarez:
I thought having a more flexible logging would help us when debugging. At the same time you want to see logs in console you like to have it stored for further reading. Anyways let's move it on.

gandarez (approved):

PR #365, opened 2021-05-06
Limit recursive calls to avoid stack overflow error

gandarez:
This PR adds a limitation to recursive calls to avoid stack overflow in some cases.


pkg/project/git.go
```diff
@@ -158,6 +157,12 @@ func findGitConfigFile(fp string, directory string, match string) (string, bool)
 		return "", false
 	}
 
+	recursiveGitCalls++
```
dron22, line 160: Can we replace the recursive call by a while loop (via for in go)? Then we do not need the global variable.

pkg/project/mercurial.go
```diff
@@ -65,6 +68,12 @@ func findHgConfigDir(fp string) (string, bool) {
 		return "", false
 	}
 
+	recursiveMercurialCalls++
```
dron22, line 71: See above

pkg/project/git.go
alanhamlett: In python we had instance variables, but a loop makes more sense in Go.

alanhamlett:
A great failsafe in case of unexpected root paths.

gandarez:
Removed recursive calls in order to have a while loop.

dron22 (approved):
Beautiful :heart:

PR #372, opened 2021-05-10
Add TFVC Project Detection

### LINKED ISSUE #75: Add TFS source control detection
As per requested by a user would be good to have Team Foundation Server detection.

gandarez:
This PR adds TFVC (TFS) project detection. It will only detect project name because branch detection is obscure and seems  you're alone in a dark room.

Also added a minor improvement by adding `defer` to some tests.

Closes #75

### DIFF
--- a/pkg/project/file.go
+++ b/pkg/project/file.go
@@ -80,7 +80,7 @@ func FindFile(fp string) (string, bool, error) {
 	return "", false, nil
 }
 
-// fileExists checks if a file exist and is not a directory.
+// fileExists checks if a file or directory exist.
 func fileExists(fp string) bool {
 	_, err := os.Stat(fp)
 	return err == nil || os.IsExist(err)
--- a/pkg/project/git.go
+++ b/pkg/project/git.go
@@ -60,16 +60,17 @@ func (g Git) Detect() (Result, bool, error) {
 	}
 
 	// Find for .git/config file
-	gitConfigFile, ok := findGitConfigFile(fp, ".git", "config")
+	gitConfigFile, ok := findFileOrDirectory(fp, ".git", "config")
 
 	if ok {
-		project := filepath.Base(filepath.Join(gitConfigFile, ".."))
+		dir, _ := filepath.Split(gitConfigFile)
+		project := filepath.Base(filepath.Join(dir, ".."))
 
-		branch, err := findGitBranch(filepath.Join(gitConfigFile, "HEAD"))
+		branch, err := findGitBranch(filepath.Join(gitConfigFile, "../HEAD"))
 		if err != nil {
 			log.Errorf(
 				"error finding for branch name from %q: %s",
-				filepath.Join(filepath.Dir(gitConfigFile), "HEAD"),
+				filepath.Join(filepath.Dir(filepath.Join(gitConfigFile, "..")), "HEAD"),
 				err,
 			)
 		}
@@ -77,18 +78,18 @@ func (g Git) Detect() (Result, bool, error) {
 		return Result{
 			Project: project,
 			Branch:  branch,
-			Folder:  filepath.Join(gitConfigFile, ".."),
+			Folder:  filepath.Join(dir, ".."),
 		}, true, nil
 	}
 
 	// Find for .git file
-	gitConfigFile, ok = findGitConfigFile(fp, "", ".git")
+	gitConfigFile, ok = findFileOrDirectory(fp, "", ".git")
 	if !ok {
 		return Result{}, false, nil
 	}
 
 	// Find for gitdir path
-	gitdir, err := findGitdir(filepath.Join(gitConfigFile, ".git"))
+	gitdir, err := findGitdir(gitConfigFile)
 	if err != nil {
 		return Result{}, false,
 			Err(fmt.Sprintf("error finding gitdir: %s", err))
@@ -123,7 +124,7 @@ func (g Git) Detect() (Result, bool, error) {
 
 	if gitdir != "" {
 		// Otherwise it's only a plain .git file
-		project := filepath.Base(gitConfigFile)
+		project := filepath.Base(filepath.Join(gitConfigFile, ".."))
 
 		branch, err := findGitBranch(filepath.Join(gitdir, "HEAD"))
 		if err != nil {
@@ -137,31 +138,11 @@ func (g Git) Detect() (Result, bool, error) {
 		return Result{
 			Project: project,
 			Branch:  branch,
-			Folder:  gitdir,
+			Folder:  filepath.Join(gitdir, ".."),
 		}, true, nil
 	}
 
 	return Result{}, false, nil
-}
-
-func findGitConfigFile(fp, dir, match string) (string, bool) {
-	i := 0
-	for i < maxRecursiveIteration {
-		if fileExists(filepath.Join(fp, dir, match)) {
-			return filepath.Join(fp, dir), true
-		}
-
-		fp = filepath.Clean(filepath.Join(fp, ".."))
-		if fp == "." || fp == "/" || driveLetterRegex.MatchString(fp) {
-			return "", false
-		}
-
-		i++
-	}
-
-	log.Warnf("didn't find git config file after %d iterations", maxRecursiveIteration)
-
-	return "", false
 }
 
 func findSubmodule(fp string, patterns []regex.Regex) (string, bool, error) {
@@ -169,12 +150,12 @@ func findSubmodule(fp string, patterns []regex.Regex) (string, bool, error) {
 		return "", false, nil
 	}
 
-	gitConfigFile, ok := findGitConfigFile(fp, "", ".git")
+	gitConfigFile, ok := findFileOrDirectory(fp, "", ".git")
 	if !ok {
 		return "", false, nil
 	}
 
-	gitdir, err := findGitdir(filepath.Join(gitConfigFile, ".git"))
+	gitdir, err := findGitdir(gitConfigFile)
 	if err != nil {
 		return "", false,
 			Err(fmt.Sprintf("error finding gitdir for submodule: %s", err))
--- a/pkg/project/git.go
+++ b/pkg/project/git.go
@@ -13,7 +13,7 @@ import (
 
 // Git contains git data.
 type Git struct {
-	// Filepath conaints the entity path.
+	// Filepath contains the entity path.
 	Filepath string
 	// SubmodulePatterns will be matched against the submodule path and if matching, will skip it.
 	SubmodulePatterns []regex.Regex
--- a/pkg/project/mercurial.go
+++ b/pkg/project/mercurial.go
@@ -12,7 +12,7 @@ import (
 
 // Mercurial contains mercurial data.
 type Mercurial struct {
-	// Filepath conaints the entity path.
+	// Filepath contains the entity path.
 	Filepath string
 }
 
@@ -31,7 +31,6 @@ func (m Mercurial) Detect() (Result, bool, error) {
 
 	// Find for .hg folder
 	hgDirectory, ok := findHgConfigDir(fp)
-
 	if ok {
 		project := filepath.Base(filepath.Join(hgDirectory, ".."))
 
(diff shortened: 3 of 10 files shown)


alanhamlett:
Very cool!

> Skipping tests for windows because it's useless as it is

Why are these tests useless on Windows OS?

gandarez:
> Why are these tests useless on Windows OS?

Because it only changes the folder name and we mock the path, so it's redundant IMO.

alanhamlett:
In CI, that's redundant but what if a dev builds on Windows then they won't run these tests until pushing to GH?

pkg/project/tfvc.go
```diff
@@ -0,0 +1,74 @@
...
+			Err(fmt.Errorf("failed to get the real path: %w", err).Error())
+	}
+
+	// Take only the directory
+	if fileExists(fp) {
+		fp = filepath.Dir(fp)
+	}
+
+	tfFolderName := ".tf"
+	if runtime.GOOS == "windows" {
```
dron22, line 33: You skip the test for windows, and here you're added windows specific behavior. Did not fully understand, why you skipped testing for windows. Could you elaborate?

pkg/project/tfvc.go
```diff
@@ -0,0 +1,74 @@
...
+	// Find for tf/properties.tf1 file
+	tfDirectory, ok := findTfConfigFile(fp, tfFolderName, "properties.tf1")
+	if ok {
+		project := filepath.Base(filepath.Join(tfDirectory, ".."))
+
+		return Result{
+			Project: project,
+			Folder:  filepath.Dir(filepath.Join(tfDirectory, "..")),
+		}, true, nil
+	}
```
dron22, line 46: Normally in go, you catch the exceptional case with a condition.
```suggestion
	if !ok {
		return Result{}, false, nil
	}
```

pkg/project/tfvc.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return Result{}, false, nil
+}
+
+func findTfConfigFile(fp, dir, match string) (string, bool) {
```
dron22, line 51: This function is not really specific to tf, right? It actually contains the same logic as `findGitConfigFile` and `findSvnConfigFile`. How about moving the function into `project.go` and use it in git/svn/tf detection?

alanhamlett:
Let's not skip this test on Windows, even if it's redundant.

pkg/project/project.go
```diff
@@ -215,6 +219,27 @@ func generateProjectName() string {
 	return strings.Join(str, " ")
 }
 
+// findFile finds for specific file until maxRecursiveIteration is reached.
+func findFile(fp, dir, match string) (string, bool) {
```
dron22, line 223: We could make input parameters names a more meaningful here and provide some more info in the comment. Some suggestion:
```suggestion
// findFile searches for a file with name `fileName`. Search starts in `startDir`
// and will traverse through all parent directories until the file is found,
// root directory is reached or `maxRecursiveIteration` is exceeded. File is
// used here in the sense of "everything is a file" and can also be a directory.
//
// If `fileDir` is specified, `fileName` is expected to be inside of a directory.
// E.g. `findFile("/some/path", ".git", "config")` will look for the git config
// file (`./.git/config`).
func findFile(startDir, fileDir, fileName string) (string, bool) {
```

dron22 (commented):
Alright nice :+1: Added 2 more small remarks

pkg/project/mercurial.go
```diff
@@ -31,26 +31,26 @@ func (m Mercurial) Detect() (Result, bool, error) {
 
 	// Find for .hg folder
 	hgDirectory, ok := findHgConfigDir(fp)
```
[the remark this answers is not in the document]
gandarez: yep we can. I actually renamed it to `findFileOrDirectory`.

pkg/project/project.go
gandarez: I'd rename it to `findFileOrDirectory` and get rid of multiple comments. Anyways I accepted your suggestion and added it.

dron22: `findFileOrDirectory` also sounds good :+1: With this name, we can actually shorten the comment. My main intention with the comments is to describe the following behavior:
- Usage of `fileDir` param
- The fact that `fileDir` is optional and can be left empty
- That the function will traverse through parent directories until break out condition is reached.
- Break out conditions
This is a lot, because this function does a lot. In go the function signature (function name + parameter names) should be meaningful and already give you information about what it does. Additionally a comment can be used to describe more details. This is very helpful, as you do not need to read the code of the function to understand it.

dron22 (approved):

gandarez:
Func has been renamed and comments revisited.

pkg/project/file.go
```diff
@@ -80,10 +80,18 @@ func FindFile(fp string) (string, bool, error) {
...
-	return err == nil || os.IsExist(err)
+	if err == nil {
+		return true
+	}
+
+	if os.IsNotExist(err) {
+		return false
+	}
+
+	return false
```
dron22, line 94: Out of interest, what was wrong with the original implementation here?

In the current form, the `os.IsNotExist(err)` condition, does not really make sense, as the two last returns are both `false`.

pkg/project/git.go
```diff
@@ -60,35 +60,35 @@ func (g Git) Detect() (Result, bool, error) {
 	}
 
 	// Find for .git/config file
-	gitConfigFile, ok := findGitConfigFile(fp, ".git", "config")
+	gitConfigFile, ok := findFileOrDirectory(fp, ".git", "config")
 
 	if ok {
-		project := filepath.Base(filepath.Join(gitConfigFile, ".."))
+		project := filepath.Base(filepath.Join(gitConfigFile, "../.."))
```
dron22, line 66: To be windows compatible, I guess this needs a small adjustment.
```
project := filepath.Base(filepath.Join(gitConfigFile, "..", ".."))
```

But in general, we're doing a little hack here, right? Because `findFileOrDirectory` will return something like `/path/to/wakatime-cli/.git/config` and then we're pretending `config` is a directory by adding `../..`. How about to strip the filename and continue with the `.git` path?

pkg/project/file.go
gandarez: The current implementation looks good though. Reverting it here.

pkg/project/git.go
gandarez: > To be windows compatible, I guess this needs a small adjustment.

Tests on windows pass 🤔 

> But in general, we're doing a little hack here, right? 

Did you mean to change the returned value?

dron22: > Did you mean to change the returned value?
What I meant w

dron22: To make it a little bit more clean, we could use `filepath.Split` (https://golang.org/pkg/path/filepath/#Split), to split the filename and then work with `/path/to/wakatime-cli/.git` in the code below.

gandarez: I see. Makes more sense though 👍

gandarez:
@dron22 fixes pushed

pkg/project/git.go
```diff
@@ -60,35 +60,36 @@ func (g Git) Detect() (Result, bool, error) {
...
-	gitConfigFile, ok := findGitConfigFile(fp, ".git", "config")
+	gitConfigFile, ok := findFileOrDirectory(fp, ".git", "config")
 
 	if ok {
-		project := filepath.Base(filepath.Join(gitConfigFile, ".."))
+		dir, _ := filepath.Split(gitConfigFile)
+		project := filepath.Base(filepath.Join(dir, ".."))
 
-		branch, err := findGitBranch(filepath.Join(gitConfigFile, "HEAD"))
+		branch, err := findGitBranch(filepath.Join(gitConfigFile, "../HEAD"))
```
dron22, line 69:
```suggestion
		branch, err := findGitBranch(filepath.Join(gitDir, "HEAD"))
```

pkg/project/git.go
```diff
@@ -60,35 +60,36 @@ func (g Git) Detect() (Result, bool, error) {
 	}
 
 	// Find for .git/config file
-	gitConfigFile, ok := findGitConfigFile(fp, ".git", "config")
+	gitConfigFile, ok := findFileOrDirectory(fp, ".git", "config")
 
 	if ok {
-		project := filepath.Base(filepath.Join(gitConfigFile, ".."))
+		dir, _ := filepath.Split(gitConfigFile)
```
dron22, line 66:
```suggestion
		gitDir := filepath.Dir(gitConfigFile)
```

pkg/project/git.go
```diff
@@ -60,35 +60,36 @@ func (g Git) Detect() (Result, bool, error) {
...
+		project := filepath.Base(filepath.Join(dir, ".."))
 
-		branch, err := findGitBranch(filepath.Join(gitConfigFile, "HEAD"))
+		branch, err := findGitBranch(filepath.Join(gitConfigFile, "../HEAD"))
 		if err != nil {
 			log.Errorf(
 				"error finding for branch name from %q: %s",
-				filepath.Join(filepath.Dir(gitConfigFile), "HEAD"),
+				filepath.Join(filepath.Dir(filepath.Join(gitConfigFile, "..")), "HEAD"),
```
dron22, line 73:
```suggestion
filepath.Join(gitDir, "HEAD"),
```

pkg/project/git.go
```diff
@@ -60,35 +60,36 @@ func (g Git) Detect() (Result, bool, error) {
 	}
 
 	// Find for .git/config file
-	gitConfigFile, ok := findGitConfigFile(fp, ".git", "config")
+	gitConfigFile, ok := findFileOrDirectory(fp, ".git", "config")
 
 	if ok {
-		project := filepath.Base(filepath.Join(gitConfigFile, ".."))
+		dir, _ := filepath.Split(gitConfigFile)
+		project := filepath.Base(filepath.Join(dir, ".."))
```
dron22, line 67:
```suggestion
		projectDir := filepath.Join(gitDir, "..")
```

pkg/project/git.go
```diff
@@ -60,35 +60,36 @@ func (g Git) Detect() (Result, bool, error) {
...
+				filepath.Join(filepath.Dir(filepath.Join(gitConfigFile, "..")), "HEAD"),
 				err,
 			)
 		}
 
 		return Result{
 			Project: project,
 			Branch:  branch,
-			Folder:  filepath.Join(gitConfigFile, ".."),
+			Folder:  filepath.Join(gitConfigFile, "../.."),
```
dron22, line 81:
```suggestion
			Project: filepath.Base(projectDir),
			Branch: branch,
			Folder:  projectDir,
```

pkg/project/git.go
```diff
@@ -137,44 +138,24 @@ func (g Git) Detect() (Result, bool, error) {
...
-	return "", false
-}
-
 func findSubmodule(fp string, patterns []regex.Regex) (string, bool, error) {
 	if !shouldTakeSubmodule(fp, patterns) {
 		return "", false, nil
 	}
 
-	gitConfigFile, ok := findGitConfigFile(fp, "", ".git")
+	gitConfigFile, ok := findFileOrDirectory(fp, "", ".git")
```
dron22, line 153: This is actually the .git directory, not the git config file.
```suggestion
	gitDir, ok := findFileOrDirectory(fp, "", ".git")
```

pkg/project/git.go
```diff
@@ -137,44 +138,24 @@ func (g Git) Detect() (Result, bool, error) {
...
 	if !ok {
 		return "", false, nil
 	}
 
-	gitdir, err := findGitdir(filepath.Join(gitConfigFile, ".git"))
+	gitdir, err := findGitdir(gitConfigFile)
```
dron22, line 158: Inside `findGitdir`, `readFile` will be executed on the input parameter. But we will pass in `.git` directory path, right? This can never work, or am I missing something?

dron22 (commented):
Added some comments regarding formatting, to use `.git` directory path throughout the code.

Also, I think I found a general issue in the git detection (see my last 2 comments). This diverges a little bit from the main topic of the PR (tfvc), but I really want these points to be addressed, as we had the recent problems in project detection logic (recursion beasts) and it is really complex and challenging to make it simple and clean.

Should we continue here, or in a different PR?

gandarez:
@dron22 issues regarding `git` will be taclked in another PR soon.

PR #381, opened 2021-05-18
Configure jww to capture logs from viper

gandarez:
This PR configures jww to capture logs from `viper`. It's useful when we need more verbose from the external lib. I also changed the order we configure logging to be the first thing to do in `run` otherwise no log is printed before it.

### DIFF
--- a/cmd/legacy/run.go
+++ b/cmd/legacy/run.go
@@ -26,18 +26,21 @@ func Run(v *viper.Viper) {
 		log.Fatalf("failed to load log params: %s", err)
 	}
 
+	logFile := os.Stdout
+
 	if !logfileParams.ToStdout {
 		log.Debugf("log to file %s", logfileParams.File)
 
-		f, err := os.OpenFile(logfileParams.File, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+		logFile, err = os.OpenFile(logfileParams.File, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
 		if err != nil {
 			log.Fatalf("error opening log file: %s", err)
 		}
 
-		log.SetOutput(f)
+		log.SetOutput(logFile)
 	}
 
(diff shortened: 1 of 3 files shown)


alanhamlett:
Awesome! This will help debug any INI parsing errors!

alanhamlett (approved):

gandarez:
> Awesome! We can also capture and log errors from the INI parsing library?

It has a debug func but the documentation says:

```
DebugFunc is called to collect debug information (currently only useful to debug parsing Python-style multiline values).
```

gandarez:
@dron22 @alanhamlett I pushed a new commit. This change enables `jww` only when debug mode is enabled (or verbose flag is true). What do you guys think? It's what Alan said "because normally we won’t want jww output".

cmd/legacy/run.go
```diff
@@ -26,18 +26,21 @@ func Run(v *viper.Viper) {
...
 		if err != nil {
 			log.Fatalf("error opening log file: %s", err)
 		}
 
-		log.SetOutput(f)
+		log.SetOutput(logFile)
 	}
 
 	log.SetVerbose(logfileParams.Verbose)
+	log.SetJww(logfileParams.Verbose, logFile)
```
dron22, line 43: `logFile` will be nil, if `logfileParams.ToStdout` equals `true`. Maybe passing `logfileParams.ToStdout` into `log.SetJww()` and handle this inside? There are different possibilities..

dron22 (changes requested):

cmd/legacy/run.go
gandarez:
```Go
func SetJww(verbose, w io.Writer) {
	if verbose {
		jww.SetLogThreshold(jww.LevelDebug)
		if w != nil {
			jww.SetLogOutput(w)
		}
	}
}
```

dron22 (approved):
Everything is a file. LGTM :+1:

PR #383, opened 2021-05-19
Improve makefile

alanhamlett:


CONTRIBUTING.md
```diff
@@ -6,14 +6,12 @@ To contribute to this project please carefully read this document.
 
 `wakatime-cli` is written in [Go](https://golang.org/).
 
-**Note:** wakatime-cli uses [Go Modules](https://github.com/golang/go/wiki/Modules) to manage dependencies.
-
 Prerequisites:
 
-- `make` - We use make to build and run tests.
+- We use make to build and run tests.
```
gandarez, line 11:
```suggestion
- We use `make` to build and run tests.
```

CONTRIBUTING.md
```diff
@@ -6,14 +6,12 @@ To contribute to this project please carefully read this document.
...
 - [Go 1.16](https://golang.org/doc/install)
 
-After cloning it first run `make test` to make sure everything is properly set.
+After cloning, install dependencies with `make install`.
```
gandarez, line 14: What about:

```
After cloning, install dependencies with `make install` and then run `make test` to make sure everything is properly set.
```

CONTRIBUTING.md
```diff
@@ -24,7 +22,7 @@ This project currently has two branches
 
 ## Testing and Linting
 
-Before any commit make sure to always run `make lint` and `make test` consecutively. It guarantees there's no linting error and all tests pass.
+Run `make test-all`.
```
gandarez, line 25: It's important to say that before pushing the PR would be better to run tests and linting to avoid failing checks at Github Actions.

gandarez (changes requested):
Can we also change change `on_push.yml` to instead calling `go mod vendor` we replace by `make install-go-modules`?

CONTRIBUTING.md
alanhamlett: The `make test` is already stated below in the testing section... I don't think it needs to be in setup.

CONTRIBUTING.md
alanhamlett: I think that's implied? If they care to check the specifics they should read the `Makefile` IMO.

alanhamlett:
All comments addressed 😄

gandarez:
Sounds good to me but we might wait until we fix semver about `misc` and `doc` branches.

gandarez (approved):
Sounds good to me but we might wait until we fix semver about `misc` and `doc` branches.

https://github.com/wakatime/semver-action/pull/15

PR #389, opened 2021-05-19
Strip quotes from parameters

gandarez:


alanhamlett (approved):

dron22 (approved):

pkg/vipertools/vipertools.go
```diff
@@ -47,7 +49,7 @@ func FirstNonEmptyString(v *viper.Viper, keys ...string) (string, bool) {
 
 	for _, key := range keys {
 		if value := v.GetString(key); value != "" {
-			return value, true
+			return strings.Trim(value, `"'`), true
```
alanhamlett, line 52: Using backticks in Go creates a string and allows using quotes inside? Pretty cool, didn't know that.

gandarez: better than `"\""`

PR #390, opened 2021-05-19
Add GetString to vipertools

gandarez:
This PR adds GetString to vipertools package and replaces the native viper function. It also strips any single or double quotes.

### DIFF
--- a/cmd/legacy/configread/configread.go
+++ b/cmd/legacy/configread/configread.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/wakatime/wakatime-cli/pkg/exitcode"
 	"github.com/wakatime/wakatime-cli/pkg/log"
+	"github.com/wakatime/wakatime-cli/pkg/vipertools"
 
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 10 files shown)


alanhamlett (commented):
Let's add a test to make sure passing `--config "\"with quotes\""` results in string `with quotes` not `"with quotes"`.

cmd/legacy/configread/configread.go
```diff
@@ -42,7 +43,7 @@ func Read(v *viper.Viper) (string, error) {
 		return "", err
 	}
 
-	value := strings.TrimSpace(v.GetString(params.ViperKey()))
+	value := strings.TrimSpace(vipertools.GetString(v, params.ViperKey()))
```
alanhamlett, line 46: We don't want to strip quotes from config values, only from command line params.

pkg/config/config_test.go
```diff
@@ -82,11 +83,15 @@ func TestFilePath(t *testing.T) {
 			ViperValue: "~/path/.wakatime.cfg",
 			Expected:   filepath.Join(home, "path", ".wakatime.cfg"),
 		},
-		"env_trailling_slash": {
+		"surrouding double quotes": {
+			ViperValue: `"~/path/.wakatime.cfg"`,
```
alanhamlett, line 87: 👍

alanhamlett (approved):

PR #394, opened 2021-05-20
Control http.Client creation instead of using http.DefaultClient

alanhamlett:
Setup `http.Client` in central location.


alanhamlett:
Fixed branch prefix in replacement PR #395.

PR #395, opened 2021-05-20
Control http.Client creation instead of using http.DefaultClient

alanhamlett:
Setup `http.Client` in central location.


dron22 (approved):
LGTM. All configurations make sense. As we do not expect, to fix the TLS bug here, it feels more like a feature though, right? :+1:

alanhamlett:
> it feels more like a feature though, right?

Yea, but possibly it would fix some default timeout issues since there was no timeout previously.

PR #401, opened 2021-05-20
Update semver to v1.3.2

gandarez:


alanhamlett:
What are the two binaries for?

alanhamlett (approved):

alanhamlett (changes requested):
Should remove the binaries that were added... I think they were added by accident?

gandarez:
yep I added them when testing and will be removed.

alanhamlett (approved):

PR #405, opened 2021-05-20
Ignore tags for on push event

gandarez:
This PR ignores tags for `on push event`. We've been running all tests twice where the first one when PR gets merged and later when tagging.

Screen Shot 2021-05-20 at 18 51 14

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -1,6 +1,11 @@
 name: "Tests"
 
-on: push
+on:
+  push:
+    branches:
+      - "**"
+    tags-ignore:
+      - "**"
 
 env:
   GO_VERSION: "1.16"


.github/workflows/on_push.yml
```diff
@@ -1,6 +1,11 @@
 name: "Tests"
 
-on: push
+on:
+  push:
+    branches:
+      - "**"
```
dron22, line 6: Isn't this the default, to be triggered for all branches? Do we need to specify this?

gandarez: I came across this note and also [here](https://github.community/t/dont-run-on-tag-creation/137469/3).
```
If you define only tags or only branches, the workflow won't run for events affecting the undefined Git ref.
```

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

dron22: Sounds good, thanks for the info :+1:

dron22 (approved):

gandarez:
This PR will be tackled by https://github.com/wakatime/wakatime-cli/pull/408

Issue #406, opened 2021-05-21
Concurrent pipelines shouldn't break semver


gandarez:
We might be aware that concurrent pipelines might break semver calculation. The explanation is simple. When a PR gets merged to either `master` or `develop` it runs (on_push.yml) 6 test jobs to aftwerwards run `Release` which is the step that the magic happens. We **MUST** merge a PR **only if no pipeline is running at develop or master**.

alanhamlett:
We can run a `Pre-Release` step that claims a version with a tag, or by creating the Draft Release?

dron22:
Claiming version is a good idea :+1: Also the tests should never fail, as we only allow PR merges with passing tests.

gandarez:
I've been working around creating a draft release but it's somehow problematic. It doesn't return correctly when calling the API also it doesn't trigger any workflow when created. Anyways I'll find a proper way to do that.

alanhamlett:
Looks like it's fixed now after #408, since these two PRs were merged around the same time and their versions came out correct. Technically there's still a race condition here, but the window is under a few seconds so it's unlikely to ever be a problem:

https://github.com/wakatime/wakatime-cli/releases/tag/v1.7.4-alpha.1
https://github.com/wakatime/wakatime-cli/releases/tag/v1.7.4-alpha.2

PR #408, opened 2021-05-22
Merge on_release workflow into on_push

### LINKED ISSUE #391: Create Releases as Draft until build assets fully uploaded
Right now we create a Release, then build and upload assets. Before the assets have uploaded, the Release is live and plugins try to download it unsuccessfully. Creating the release as Draft, and removing Draft boolean after uploading assets will fix this.

gandarez:
This PR unifies `on release` workflow and `on push`. It solves the problem where creating a release triggers the workflow to build binaries which takes around 10 min to complete making the plugins fail because there's no asset to be downloaded. By unifying both workflows we might expect a new release to be created only when all binaries are built and uploaded.

Screen Shot 2021-05-22 at 09 17 52

It also includes the changes at https://github.com/wakatime/wakatime-cli/pull/405
Closes https://github.com/wakatime/wakatime-cli/issues/391


alanhamlett:
Related to #406 too?

alanhamlett (approved):

gandarez:
> Related to #406 too?

Concurrency might be fixed with this PR but we might validate that by running few more pipelines.

PR #409, opened 2021-05-22
Prevent using DefaultTransport

alanhamlett:
This adds to #395 to make sure `NewTransport` is used instead of `DefaultTransport`.


pkg/api/transport.go
```diff
@@ -16,3 +16,12 @@ func NewTransport() *http.Transport {
 		ForceAttemptHTTP2:   true,
 	}
 }
+
+// GetOrCreateTransport gets the client's Transport if already exists, or initializes a new one.
+func GetOrCreateTransport(c *Client) *http.Transport {
```
gandarez: What if we pass a pointer to http.Transport to `NewTransport` and then do something like this:

```Go
func NewTransport(t *http.Transport) *http.Transport {
	if t != nil {
		return t.Clone()
	}

	return &http.Transport{
		Proxy:               nil,
		TLSHandshakeTimeout: 30 * time.Second,
		MaxIdleConns:        1,
		MaxIdleConnsPerHost: 1,
		MaxConnsPerHost:     1,
		ForceAttemptHTTP2:   true,
	}
}
```

Does that make sense @dron22 @alanhamlett?

alanhamlett: Sure, just combined the two functions into 1.

dron22 (changes requested):
If we use `NewClient`, as it was adjusted in https://github.com/wakatime/wakatime-cli/pull/409, the client should always have the custom transport set. Thus default transport will never be used in the option functions and the changes here should not have any effect.

@alanhamlett Am I missing something?

alanhamlett:
Yea, I saw the same thing but `client.Transport` is nullable and it doesn't hurt to be extra sure.

dron22:
If we worry, that transport is unintentionally set to `nil`, we could also worry that transport is unintentionally set to DefaultTransport somewhere. :thinking: 

But alright, if you prefer double security, then let's do it. We do not need to change the signature of `NewTransport` to achieve this though. The previous implementation was simpler and using it as-is will make the reasoning about the option functions easier.

alanhamlett:
> The previous implementation was simpler

The one where I used a `GetOrCreateTransport` function?

dron22:
I meant the version before the changes in this PR (`NewTransport()`), which had no input arguments, and simply returned the custom transport. The new name `HTTPTransport` is also good, I mainly care about the input argument(s).

Reason: Passing the http client as an input argument makes it harder to reason about the function. Instead of just looking at the function signature, you need to read the implementation to be sure about what it is doing. Also the option of passing `nil` is handled, which adds more complexity.

gandarez:
Is this PR still relevant?

alanhamlett:
> Is this PR still relevant?

I think so... we want to remove any coding calling `http.DefaultTransport`.

pkg/api/transport.go
```diff
@@ -6,8 +6,12 @@ import (
 	"time"
 )
 
-// NewTransport creates a new http.Transport, with default 30 second TLS timeout.
-func NewTransport() *http.Transport {
```
dron22, line 10: Let's keep the original signature unchanged here. We can achieve the same without changing it. As commented above, this will make the logic simpler, as we do not manipulate a passed in object and have no special case for passing in nil.

dron22 (changes requested):
Referring to my comment above here.

pkg/api/transport.go
alanhamlett: Updated.

PR #411, opened 2021-05-23
Remove gocertifi

### LINKED ISSUE #385: x509: certificate signed by unknown authority
checked every file and reinstalled the extension for about 100 times and regenerated API key also but nothing is working.
Please Help

alanhamlett:
Manually bundle LetsEncrypt certs instead of using gocertifi, and also use system ca certs.

Using the system's cacerts might fix #385. I didn't want to use anything external from the system before, but this looks safe and might be necessary for proxies.


dron22 (approved):
Alright, this sounds like a valid idea to solve the issues :+1:

Issue #404, opened 2021-05-20
Should not send heartbeat for null entity


alanhamlett:
When executing without `--entity` and only `--sync-offline-activity`, it's parsing the current working directory as the entity. Instead, it should detect null as the entity and only sync offline heartbeats, skipping sending entity.

```
{"caller":"/root/wakatime-cli/pkg/filestats/filestats.go:68","func":"1","level":"warning","message":"failed to detect the total number of lines in file \"/root/wakatime-cli\": read /root/wakatime-cli: is a directory","now":"2021-05-20T10:47:14-07:00","version":"<local-build>"}
{"caller":"/root/wakatime-cli/pkg/language/language.go:41","func":"1","level":"warning","message":"failed to detect language on file entity \"/root/wakatime-cli\": could not detect the language of file \"/root/wakatime-cli\"","now":"2021-05-20T10:47:14-07:00","version":"<local-build>"}
```

PR #414, opened 2021-05-24
Move api client initialization into api package

dron22:
This PR move the initialization of the api client into a function `NewClientFromParams` in `api` package. This removes duplicate code, as the api client was initialized in various commands (`heartbeat`, `today` and `todaygoal`). It is a preparation for implementation of sync offline command (https://github.com/wakatime/wakatime-cli/issues/404).

- Api client initialization is moved into `api.NewClientFromParams`. Main initialization logic is only moved and not modified.
- `legacyparams.API` and `legacyparams.Network` are unified into `api.Params`.
- Tests for api related parameter loading from viper is unified in `legacyparams` package and removed everywhere else. 
- `sync-offline-activity` command will be able to use the initialization logic via `api.NewClientFromParams` and does not need to implement it itself.


alanhamlett:
Branch is outdated, needs a rebase.

dron22:
> Branch is outdated, needs a rebase.

Thanks for the heads-up :+1:

gandarez (approved):
LGTM regardless the name `NewClientFromParams()` we've been discussing on Slack.

cmd/legacy/legacyparams/params.go
```diff
@@ -32,7 +32,33 @@ var (
...
+	ProxyURL         string
+	SSLCertFilepath  string
+	Timeout          time.Duration
+	URL              string
+}
+
+// String implements fmt.Stringer interface.
+func (p API) String() string {
+	return fmt.Sprintf(
+		"api key: '%s', api url: '%s', plugin: '%s', timeout: %s, disable ssl verify: %t,"+
```
gandarez, line 52: Isn't it better to surround strings with double quote instead single quote?

dron22: Valid question. Actually using `%q` placeholder would be simpler, yes. But the reason I put them is that in the json log output, the message itself will be surrounded with double quotes as well. Consequently we would end up with escaped double quotes, which is slightly harder to read.
Example:
```
{"message":"api key: \"00635xxx\", api url: \"https://wakatime.com\", plugin: \"WakaTime/1.23\""}
```
versus:
```
{"message":"api key: '00635xxx', api url: 'https://wakatime.com', plugin: 'WakaTime/1.23'"}
```

gandarez (approved):

PR #420, opened 2021-05-26
Publish releases to prevent leaving as Drafts

alanhamlett:
Releases are currently left as Draft, hopefully this fixes it.


alanhamlett:
It worked! Not sure why releases were left as Drafts when the release action's docs say `draft` defaults to `false`.

https://github.com/softprops/action-gh-release/blob/9729932bfb75c05ad1f6e3a729294e05abaa7001/action.yml#L19

PR #422, opened 2021-05-27
Validate file entity in heartbeat command

### LINKED ISSUE #403: Inconsistent behavior when passing invalid file with --enitity
I'm noticing two different log messages when passing `--entity fakefile`. It should always be consistent with a message along the lines of `File does not exist; ignoring this heartbeat.`

```
"caller":"wakatime-cli/cmd/legacy/heartbeat/heartbeat.go:49","func":"Run","level":"fatal","message":"failed to send heartbeat(s): failed to load command parameters: failed to retrieve entity"

"caller":"wakatime-cli/pkg/heartbeat/heartbeat.go:65","func":"New","level":"warning","message":"failed to resolve the real path of \"/user/fakefile\": lstat /root/fakefile: no such file or directory"
```

Found from https://github.com/wakatime/wakatime/blob/master/tests/test_arguments.py#L198.

dron22:
This PR adds validation of the passed in file entity to the heartbeat command. In case the file is not existing, it will exit after logging the following message:

```
{"caller":"/home/don/code/go/src/github.com/wakatime/wakatime-cli/cmd/legacy/heartbeat/heartbeat.go:47","func":"Run","level":"fatal","message":"failed to send heartbeat(s): failed to load command parameters: file 'xxx' does not exist. ignoring this heartbeat","now":"2021-05-27T23:48:48+02:00","version":"v0.0.1-test"}
```

Same is applied to extra heartbeats as well. In case a file for one extra heartbeat is not existing, execution will be aborted. @alanhamlett Does this behavior fit, what you had in mind? 

Closes #403


alanhamlett (approved):
We should also consider remote files over network drives in the future.

dron22:
> We should also consider remote files over network drives in the future.

@alanhamlett Sounds good. You will add a separate ticket for this, right?

alanhamlett (approved):

alanhamlett:
Yep, created issue for it #426.

PR #421, opened 2021-05-27
Add distinct sync-offline-activity command

dron22:
This PR adds a distinct command for syncing offline activity to handle flag `--sync-offline-activity` correctly.

Should only be merged, after merge of https://github.com/wakatime/wakatime-cli/pull/414. Ready for review though.

Closes #404

PR #423, opened 2021-05-28
Check for required args passed in

### LINKED ISSUE #402: Print --help contents when called without any arguments
We need at least one of these present, or exit with error code and print `--help` contents:

* --entity
* --today
* --today-goal
* --help
* --version
* --sync-offline-activity

Found from https://github.com/wakatime/wakatime/blob/master/tests/test_arguments.py#L218

gandarez:
This PR checks if one of required args are passed in to the binary. One of the following args will be mandatory:

- entity
- today
- today-goal
- version
- sync-offline-activity
- help

Fixes https://github.com/wakatime/wakatime-cli/issues/402

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"os"
+	"strings"
 
 	"github.com/wakatime/wakatime-cli/cmd/legacy"
 	"github.com/wakatime/wakatime-cli/pkg/exitcode"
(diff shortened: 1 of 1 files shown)


gandarez:
Converted to draft because integration tests has broken.

cmd/root.go
```diff
@@ -28,6 +40,10 @@ func NewRootCMD() *cobra.Command {
 		Use:   "wakatime-cli",
 		Short: "Command line interface used by all WakaTime text editor plugins.",
 		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) == 0 || !isOneOfRequiredArgsFilled(args) {
```
alanhamlett: Let's move `len(args) == 0` into function `isOneOfRequiredArgsFilled`

dron22 (commented):
Actually my code changes in https://github.com/wakatime/wakatime-cli/pull/421 already take care of this. It handles this directly in `./cmd/legacy/run.go`. Advantage is, that we do not have to maintain it in an additional place.

gandarez:
> Actually my code changes in #421 already take care of this. It handles this directly in `./cmd/legacy/run.go`. Advantage is, that we do not have to maintain it in an additional place.

Checking for required args in `PreRun` event has advantages. It runs before any code and we don't need to go through all those ifs before actually checking it.

cmd/root.go
gandarez: Fixed.

alanhamlett:
I really like @dron22's `log.Fatal` solution in #421 because we can add/remove required params there without having to update a list of required params in another file.

alanhamlett:
Closing in favor of #421.

PR #425, opened 2021-05-28
Calculate semver version in build job

### LINKED ISSUE #424: Include wakatime-cli version in every log line output

dron22:
Since building of binaries is triggered by `push` event now, we cannot take the version from the release event, but have to run semver action also at in build job to calculate it.

I decided to run semver a second time, as running it once and sharing data between jobs is more complicated. As jobs run on different runners, variables cannot be easily shared (https://github.community/t/sharing-a-variable-between-jobs/16967). We could write it to a file and upload/download as an artifact, but as semver action has a number of outputs, we would need to share these as well.

Closes #424

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -158,9 +158,16 @@ jobs:
         with:
           go-version: "1.16"
       -
+        name: "Calculate semver tag"
+        id: semver-tag
+        uses: wakatime/semver-action@v1.3.2
+        with:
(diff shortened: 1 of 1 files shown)


alanhamlett:
Getting the version at build time means two PRs merged at the same time might claim the same version, with one failing because the tag already exists.

This means we need to create the tag before building to claim a version.

alanhamlett:
Let's use #427 instead.

PR #427, opened 2021-05-29
Use workflow job outputs to preserve semver between jobs

### LINKED ISSUE #424: Include wakatime-cli version in every log line output

alanhamlett:
Adds a new job `Tag` doing:

* calculate semver after tests pass
* calculate changelog
* create annotated tag for new version
* add version and other info to job output, so other jobs can use without re-running semver-action

Creating the tag "reserves" a version number to prevent two PRs merged around the same time from trying to use the same version.

Closes #424

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -143,11 +143,52 @@ jobs:
         name: "Integration tests"
         run: make test-integration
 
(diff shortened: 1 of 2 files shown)


alanhamlett:
Closing in favor of #428.

PR #428, opened 2021-05-29
Add concurrency attribute to avoid claiming the same version

### LINKED ISSUE #424: Include wakatime-cli version in every log line output

gandarez:
This PR suggests few changes to release pipeline.

- Adds attribute `concurrency` to avoid claiming the same version by two concurrent jobs. Docs can be found [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency).
- Creates a separate job (called `version`) for claiming semver tag and outputs (shares) the result to other jobs. It won't outputs changelog at this step since it's used only at `release` job.
- Makes `build`, `sign` and `release` dependant on `version`.
- Adds a fallback to delete current tag if any jobs fail.

Related to https://github.com/wakatime/wakatime-cli/pull/427
Fixes https://github.com/wakatime/wakatime-cli/issues/424

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -143,11 +143,52 @@ jobs:
         name: "Integration tests"
         run: make test-integration
 
+  version:
+    name: Version
+    concurrency: tagging
+    if: ${{ github.ref == 'refs/heads/release' || github.ref == 'refs/heads/develop' }}
+    runs-on: ubuntu-latest
+    needs: [
+      test,
+      test-integration,
+      test-windows,
+      test-integration-windows,
+      test-macos,
+      test-integration-macos]
+    outputs:
+      semver_tag: ${{ steps.semver-tag.outputs.semver_tag }}
+      ancestor_tag: ${{ steps.semver-tag.outputs.ancestor_tag }}
+      is_prerelease: ${{ steps.semver-tag.outputs.is_prerelease }}
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Calculate semver tag
+        id: semver-tag
+        uses: wakatime/semver-action@v1.3.2
+        with:
+          prerelease_id: alpha
+          main_branch_name: release
+      - name: Create tag
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ github.token }}
+          script: |
+            github.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: "refs/tags/${{ steps.semver-tag.outputs.semver_tag }}",
+              sha: context.sha
+            })
+
   build:
     name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/on_push.yml
```diff
@@ -143,11 +143,52 @@ jobs:
         name: "Integration tests"
         run: make test-integration
 
+  version:
+    name: Version
+    concurrency: tagging
+    if: ${{ github.ref == 'refs/heads/release' || github.ref == 'refs/heads/develop' }}
+    runs-on: ubuntu-latest
+    needs: [
```
alanhamlett, line 151: I like it all on the same line, since we can't use trailing commas here

.github/workflows/on_push.yml
```diff
@@ -143,11 +143,52 @@ jobs:
...
+        uses: wakatime/semver-action@v1.3.2
+        with:
+          prerelease_id: alpha
+          main_branch_name: release
+      - name: Create tag
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ github.token }}
+          script: |
+            github.git.createRef({
```
alanhamlett, line 180: This creates a lightweight tag, but we've been creating an annotated tag indirectly via Releases. Let's use `github.git.createTag` like https://github.com/wakatime/wakatime-cli/pull/427/files#diff-53dbc06610f897f66670a951aca3d54295c4a0fa85d7869295330aa05265a761R179?

alanhamlett: Unless creating the release "upgrades" the lightweight tag to an annotated tag?

alanhamlett (approved):
looks good!

.github/workflows/on_push.yml
gandarez: My intention here was to make it more readable since the line was too long.

alanhamlett: Is this a YAML list? If so, we can remove `[,]` and just put each dependency on it's own line without commas.

alanhamlett: From [the docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) it's not clear if it must be a string or can be a YAML list.

.github/workflows/on_push.yml
gandarez: None of our tags are annotated. Running the following command tells if tag is annotated or not. `git for-each-ref refs/tags`:
Screen Shot 2021-05-29 at 18 33 03

Docs says:
```
[T]he output has three fields: The hash of an object, the type of the object, and the name in refs/tags that refers to the object. A so-called "lightweight" tag is a name in refs/tags that refers to a commit¹ object. An "annotated" tag is a name in refs/tags that refers to a tag object.
```
Do you want to make then annotated instead?

alanhamlett: No, let's use lightweight tags. My mistake.

PR #430, opened 2021-05-30
Get correct date string in makefile

gandarez:
This PR fixes the date argument passed to version package. It includes the correct timezone info.

current: `2021-05-30T14:44:45Z`
with fix: `2021-05-30T14:44:45 UTC`

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -243,7 +243,7 @@ func TestVersionVerbose(t *testing.T) {
 	out := run(exec.Command(binaryPath(t), "--version", "--verbose"))
 
 	assert.Regexp(t, regexp.MustCompile(fmt.Sprintf(
-		"wakatime-cli\n  Version: v0.0.1-test\n  Commit: [0-9a-f]{7}\n  Built: [0-9-:TZ]{20}\n  OS/Arch: %s/%s\n",
+		"wakatime-cli\n  Version: v0.0.1-test\n  Commit: [0-9a-f]{7}\n  Built: [0-9-:T]{19} UTC\n  OS/Arch: %s/%s\n",
 		runtime.GOOS,
 		runtime.GOARCH,
 	)), out)
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,12 @@
 BINARY_NAME?=wakatime-cli
 BUILD_DIR?="./build"
 COMMIT?=$(shell git rev-parse --short HEAD)
-DATE?=$(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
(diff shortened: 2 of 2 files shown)


main_test.go
```diff
@@ -243,7 +243,7 @@ func TestVersionVerbose(t *testing.T) {
 	out := run(exec.Command(binaryPath(t), "--version", "--verbose"))
 
 	assert.Regexp(t, regexp.MustCompile(fmt.Sprintf(
-		"wakatime-cli\n  Version: v0.0.1-test\n  Commit: [0-9a-f]{7}\n  Built: [0-9-:TZ]{20}\n  OS/Arch: %s/%s\n",
+		"wakatime-cli\n  Version: v0.0.1-test\n  Commit: [0-9a-f]{7}\n  Built: [0-9-:T]{19}UTC\n  OS/Arch: %s/%s\n",
```
alanhamlett: does this mean running this test on a machine not set to UTC will fail?

gandarez: My machine is GMT-3 and it doesn't matter. The date value is taken by running `date -u '+%Y-%m-%dT%H:%M:%SZ'` and the `-u` flag prints in UTC.

alanhamlett (approved):

dron22 (commented):
Can you explain, which problem is fixed here? Asking, because UTC and Zulu time are interchangeable and mean the same.

Also RFC3339 and ISO8601 both use the Z designator:
- https://datatracker.ietf.org/doc/html/rfc3339
- https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC

gandarez:
> Can you explain, which problem is fixed here? Asking, because UTC and Zulu time are interchangeable and mean the same.
> 
> Also RFC3339 and ISO8601 both use the Z designator:
> 
> * https://datatracker.ietf.org/doc/html/rfc3339
> * https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC

The letter `Z` to represent universal time is rarely used compared to `UTC`. I tried to make it cleaner for anyone which is not used to `Z` only.

alanhamlett:
In ISO8601, `Z` is pretty common. If it's not broken, let's not fix it.

alanhamlett:
Just to make sure, is this for displaying to a user? If so, we don't need ISO8601 format and we should add a space before `UTC` for readability.

If it's not for displaying to users, let's close without changing.

gandarez:
> Just to make sure, is this for displaying to a user? If so, we don't need ISO8601 format and we should add a space before `UTC` for readability.
> 
> If it's not for displaying to users, let's close without changing.

Yes, it's used only for `--version` flag.

alanhamlett (approved):

PR #434, opened 2021-05-31
Show help as fallback

### LINKED ISSUE #402: Print --help contents when called without any arguments

gandarez:
This PR prints help when no args are set.

Closes https://github.com/wakatime/wakatime-cli/issues/402


dron22 (approved):
very HELPful :+1:

PR #436, opened 2021-05-31
Send diagnostic info on heartbeat, today and todaygoal commands

### LINKED ISSUE #416: Send diagnostic info on error
Send diagnostic info to the backend, whenever an error occurs.

- Should send with ssl verification disabled to be able to send diagnostics on ssl verification problems.

dron22:
This PR adds diagnostics sending to the command execution.

- Will send diagnostics on every error happening on command execution.
- Diagnostics include logs and stacktrace.
- On failure, will retry with ssl verification disabled.

To achieve this, command execution is unified (includes removal of unnecessary custom errors) and wrapped into `runCmd`. This captures logs, catches panics, send the diagnostics on any potential error and exits with the specific exit code.

**Outlook**

 Sending TLS specifics was not included yet and will be addressed later.

Closes #416


pkg/api/diagnostic_test.go
```diff
@@ -0,0 +1,61 @@
...
+		err := req.ParseForm()
+		require.NoError(t, err)
+
+		assert.Equal(t, map[string][]string{
+			"platform":      {"linux"},
+			"architecture":  {"amd64"},
+			"editor":        {"vim"},
+			"error":         {"some error"},
+			"logs":          {"some logs"},
+			"stack":         {"some stack"},
```
alanhamlett: named `stacktrace` in api

pkg/api/diagnostic.go
```diff
@@ -0,0 +1,71 @@
...
+	data.Set("architecture", version.Arch)
+	data.Set("editor", plugin)
+	data.Set("error", errorMsg)
+
+	for _, d := range diagnostics {
+		switch d.Type {
+		case diagnostic.TypeLogs:
+			data.Set("logs", d.Value)
+		case diagnostic.TypeStack:
+			data.Set("stack", d.Value)
```
alanhamlett: named `stacktrace`

alanhamlett (changes requested):
All the changes look good, but we need to disable SSL cert verification when sending diagnostics otherwise we won't get the x509 errors reported.

dron22:
> All the changes look good, but we need to disable SSL cert verification when sending diagnostics otherwise we won't get the x509 errors reported.

Will add this :+1:

dron22:
@alanhamlett I added retry with ssl verification disabled

cmd/legacy/run.go
```diff
@@ -125,3 +124,77 @@ func Run(v *viper.Viper) {
...
+	params, err := legacyparams.Load(v)
+	if err != nil {
+		log.Errorf("failed to load parameters for sending diagnostics: %s", err)
+
+		return
+	}
+
+	c, err := legacyapi.NewClient(params.API)
+	if err != nil {
+		log.Errorf("failed initialize api client for sending diagnostics: %s", err)
```
gandarez:
```suggestion
		log.Errorf("failed to initialize api client for sending diagnostics: %s", err)
```

pkg/diagnostic/diagnostic.go
```diff
@@ -0,0 +1,45 @@
...
+
+const (
+	// TypeUnknown designates an unknown type of diagnostic.
+	TypeUnknown Type = iota
+	// TypeLogs designates a Diagnostic for logs.
+	TypeLogs
+	// TypeStack designates a Diagnostic for stack trace.
+	TypeStack
+	// TypeTLS designates a Diagnostic for tls logs.
+	TypeTLS
```
gandarez: Is that a preparation for something? It's only used for testing purposes.

dron22: Yes, it is the preparation for sending TLS logs. Do you think we should remove it from this PR?

gandarez: If any error during the commands execution will be caught, do we really need this type? Aren't `logs` and `stack` enough? As far I understood this type will be for specific SSL errors. Can't we always send them as "generic"?

dron22: @gandarez To be honest, I did not find a way yet to retrieve these tls logs. I was playing around with capturing stdout and stderr, but this is a little tricky and also no tls logs showed up. Even when I set `http1debug=1` via `GODEBUG`. I will remove `TypeTLS` and then we can have a discussion on it outside of this PR. :+1:

gandarez (approved):
Let's get it working ⚒️

pkg/api/diagnostic.go
```diff
@@ -0,0 +1,68 @@
...
+			return fmt.Errorf("Unknown diagnostic type %d", d.Type)
+		}
+	}
+
+	req, err := http.NewRequest(http.MethodPost, url, strings.NewReader(data.Encode()))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %s", err)
+	}
+
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
gandarez: We might set `application/json` here as discussed [here](https://wakatime.slack.com/archives/GUWD67UE4/p1622765826010600).

gandarez (commented):
one small remark.

pkg/api/diagnostic.go
dron22: I adjusted it :+1:  Now we send a json request body.

pkg/api/diagnostic_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	router.HandleFunc("/v1/plugins/errors", func(w http.ResponseWriter, req *http.Request) {
+		numCalls++
+
+		// check method and headers
+		assert.Equal(t, http.MethodPost, req.Method)
+		assert.Equal(t, []string{"application/json"}, req.Header["Content-Type"])
+
+		// check body
+		expectedBody, err := ioutil.ReadFile("testdata/diagnostics_request.json")
```
alanhamlett, line 32: should we test to make sure `Authentication` header isn't present?

dron22: Makes sense, but the test in `api` package initializes the api client itself. Only a test in `./cmd/legacy` will give full assurance, as this is the place, where the client is initialized in real execution. Will add it :+1:

dron22: @alanhamlett Check latest commit, which disables auth for diagnostics and verifies it with a test.

alanhamlett (approved):

PR #441, opened 2021-06-03
Expand users home dir for input flags

### LINKED ISSUE #440: Expand user's home dir when reading flags

gandarez:
This PR adds a feature to expand user's home directory of all input flag that are file paths. It prevents errors if paths starts with "~".

Closes https://github.com/wakatime/wakatime-cli/issues/440

### DIFF
--- a/pkg/api/option.go
+++ b/pkg/api/option.go
@@ -13,7 +13,6 @@ import (
 	"github.com/wakatime/wakatime-cli/pkg/log"
 
 	"github.com/Azure/go-ntlmssp"
-	"github.com/mitchellh/go-homedir"
 )
 
 // Option is a functional option for Client.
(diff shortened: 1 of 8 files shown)


pkg/api/option.go
```diff
@@ -130,12 +129,7 @@ func WithProxy(proxyURL string) (Option, error) {
 
 // WithSSLCertFile overrides the default CA certs file to trust specified cert file.
 func WithSSLCertFile(filepath string) (Option, error) {
-	expanded, err := homedir.Expand(filepath)
-	if err != nil {
-		return nil, fmt.Errorf("failed expanding filepath %q: %s", filepath, err)
-	}
-
-	caCert, err := ioutil.ReadFile(expanded)
+	caCert, err := ioutil.ReadFile(filepath)
```
dron22, line 132: :+1:

dron22 (changes requested):
This looks good to me :+1: Additionally, we need to expand it for extra heartbeats as well.

pkg/api/option.go
alanhamlett: We should also expand it for `--config` and `--log-file`.

gandarez: > We should also expand it for `--config` and `--log-file`.

Them are already being expanded.

gandarez:
I removed dedicated tests for tilde and moved to ones already implemented to simplify it. Also added expansion to extra heartbeats.

alanhamlett (approved):

gandarez:
@dron22 last approval.

dron22 (approved):

PR #443, opened 2021-06-04
Fix changelog for Slack notification

### LINKED ISSUE #442: Clean changelog before sending to Slack
Sometimes we add crasis to commit message and it breaks the Slack action. Before actually sending to Slack the changelog might be cleaned out.

https://github.com/wakatime/wakatime-cli/runs/2740740701
Screen Shot 2021-06-03 at 21 37 29

gandarez:
This PR replaces single quotes and crasis when found to avoid breaking Slack notification.

Fixes https://github.com/wakatime/wakatime-cli/issues/442

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -490,16 +490,26 @@ jobs:
           exclude: |
             ^Merge pull request .*
       -
+        if: ${{ github.ref == 'refs/heads/develop' }}
+        name: Prepare changelog (develop)
+        id: changelog-develop
+        shell: bash
+        run: |
+          changelog=${{ steps.changelog.outputs.changelog }}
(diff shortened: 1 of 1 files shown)


dron22 (commented):
As the workflow is growing in size and complexity, how about moving the changelog logic into a bash script in `bin` folder, to decrease size of the workflow?

gandarez:
> As the workflow is growing in size and complexity, how about moving the changelog logic into a bash script in `bin` folder, to decrease size of the workflow?

I think the size of the workflow doesn't matter and breaking into smaller parts make it harder to read and understand at once. Having a single file for that is commonly used across projects.

alanhamlett:
Let's get this one merged and discuss a separate script in another issue/Slack.

alanhamlett (approved):

PR #444, opened 2021-06-04
Lower default offline sync level

dron22:
This PR lowers the default offline sync limit to 24. This way the main heartbeat, plus 24 potential offline heartbeats, will result in 25 heartbeats being sent to backend. This matches the maximum number of heartbeats, the backend will accept at once. 

Slack conversation: https://wakatime.slack.com/archives/GUWD67UE4/p1622674192009000?thread_ts=1622669089.008400&cid=GUWD67UE4

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -16,7 +16,7 @@ const (
 	defaultTimeoutSecs = 60
 	// defaultOfflineSync is the default maximum number of heartbeats from the
 	// offline queue, which will be synced upon sending heartbeats to the API.
-	defaultOfflineSync = "100"
+	defaultOfflineSync = "24"
 )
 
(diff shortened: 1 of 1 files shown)


alanhamlett (approved):

gandarez:
What will happen if there's 1000 heartbeats in the queue? For every new heartbeat it will send 24 stored in offline queue?

alanhamlett:
The Python wakatime-cli would send 24 per request, in a loop so if there's 1000 heartbeats in offline db it would make 42 requests.

alanhamlett (approved):

alanhamlett:
Oh, looks like this just sets the default argument for `--offline-sync-activity`. We actually want that default to be high, like 1000. What we need is to never send more than 24 heartbeats per http request.

PR #451, opened 2021-06-09
Support legacy api_url param format by removing endpoint from url

### LINKED ISSUE #384: New api_base_url config, keeping legacy api_url config unchanged for backwards compatibility
Someone just brought up this issue https://github.com/muety/wakapi/issues/203 in Wakapi. It looks like the format of the `api_url` as configured in `~/.wakatime.cfg` has changed and is now supposed to be `/api` instead of `/api/heartbeat`, is that correct?

This is not particularly a bug report, but rather I'm looking for ideas on how to gracefully migrate existing users' setups, who use custom `api_url`s, to this new version.

alanhamlett:
Fixes #384.

### DIFF
--- a/cmd/legacy/legacyparams/params.go
+++ b/cmd/legacy/legacyparams/params.go
@@ -123,6 +123,13 @@ func loadAPIParams(v *viper.Viper) (API, error) {
 		apiURL = u
 	}
 
+	// remove endpoint from api base url to support legacy api_url param
+	apiURL = strings.TrimSuffix(apiURL, "/")
+	apiURL = strings.TrimSuffix(apiURL, ".bulk")
+	apiURL = strings.TrimSuffix(apiURL, "/users/current/heartbeats")
+	apiURL = strings.TrimSuffix(apiURL, "/heartbeats")
+	apiURL = strings.TrimSuffix(apiURL, "/heartbeat")
+
 	proxyURL, _ := vipertools.FirstNonEmptyString(v, "proxy", "settings.proxy")
 
 	rgx := proxyRegex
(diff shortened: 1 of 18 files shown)


cmd/root.go
```diff
@@ -41,8 +41,10 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
 	flags := cmd.Flags()
 	flags.String("alternate-language", "", "Optional alternate language name. Auto-detected language takes priority.")
 	flags.String("alternate-project", "", "Optional alternate project name. Auto-detected project takes priority.")
-	flags.String("api-url", "", "Heartbeats api url. For debugging with a local server.")
-	flags.String("apiurl", "", "(deprecated) Heartbeats api url. For debugging with a local server.")
+	flags.String("api-url", "", "API base url used for sending heartbeats and fetching code stats. Defaults to "+
```
gandarez: When the line is too long we usually break each argument into new line.
```suggestion
	flags.String(
	    "api-url",
	    "",
	    "API base url used for sending heartbeats and fetching code stats. Defaults to https://api.wakatime.com/api/v1/."
```

cmd/legacy/legacyparams/params.go
```diff
@@ -123,6 +123,13 @@ func loadAPIParams(v *viper.Viper) (API, error) {
 		apiURL = u
 	}
 
+	// remove endpoint from api base url to support legacy api_url param
+	apiURL = strings.TrimSuffix(apiURL, "/")
+	apiURL = strings.TrimSuffix(apiURL, ".bulk")
+	apiURL = strings.TrimSuffix(apiURL, "/users/current/heartbeats")
+	apiURL = strings.TrimSuffix(apiURL, "/heartbeats")
+	apiURL = strings.TrimSuffix(apiURL, "/")
```
gandarez: is that a duplicate of line 127?

gandarez (changes requested):

gandarez (approved):
LGTM

Issue #449, opened 2021-06-07
Send all queued heartbeats at once


dron22:
On heartbeat command execution, all previously queued heartbeat should be sent to WakaTime Api.

- In case the total number of queued heartbeats is greater than the offline sync limit, multiple requests should be sent.
- Should also be executed, if `offline-disabled` is set.

PR #453, opened 2021-06-11
Distinct offline sync function (send all queued preparation pt1)

dron22:
This PR adds a distinct function `offline.Sync()` for offline activity syncing. This is a first step, to separate offline syncing from the heartbeats pipeline. It is a preparation to sync all queued heartbeats on wakatime cli execution.

- Adds helper functions `popHeartbeats` and `pushHeartbeats` to pop and push without keeping the db connection open.

Related to: https://github.com/wakatime/wakatime-cli/issues/449

### DIFF
--- a/pkg/offline/offline.go
+++ b/pkg/offline/offline.go
@@ -185,32 +185,9 @@ func Sync(filepath string, syncLimit int) func(next heartbeat.Handle) error {
 
 		results, err := next(hh)
 		if err != nil {
-			var count int
-
-			for {
-				if count >= maxRequeueAttempts {
-					log.Errorf("abort requeuing after %d unsuccessful attempts", count)
-					break
-				}
-
-				requeueErr := pushHeartbeats(filepath, hh)
-				if requeueErr != nil {
-					data, jsonErr := json.Marshal(hh)
-					if jsonErr != nil {
-						log.Warnf("failed to json marshal heartbeats: %s. heartbeats: %#v", jsonErr, hh)
-					}
-
-					log.Warnf("failed to push heatbeats after api error to queue: %s. heartbeats: %s", requeueErr, string(data))
-					count++
-
-					sleepSeconds := math.Pow(2, float64(count))
-
-					time.Sleep(time.Duration(sleepSeconds) * time.Second)
-
-					continue
-				}
-
-				break
+			requeueErr := pushHeartbeatsWithRetry(filepath, hh)
+			if requeueErr != nil {
+				log.Warnf("failed to push heatbeats to queue after api error: %s", requeueErr)
 			}
 
 			return err
@@ -298,6 +275,39 @@ func popHeartbeats(filepath string, limit int) ([]heartbeat.Heartbeat, error) {
 	}
 
 	return queued, nil
+}
+
+func pushHeartbeatsWithRetry(filepath string, hh []heartbeat.Heartbeat) error {
+	var (
+		count int
+		err   error
+	)
+
+	for {
+		if count >= maxRequeueAttempts {
+			data, jsonErr := json.Marshal(hh)
+			if jsonErr != nil {
+				log.Warnf("failed to json marshal heartbeats: %s. heartbeats: %#v", jsonErr, hh)
+			}
+
+			return fmt.Errorf("abort requeuing after %d unsuccessful attempts: %s. heartbeats: %s", count, err, string(data))
+		}
+
+		err = pushHeartbeats(filepath, hh)
+		if err != nil {
+			count++
+
+			sleepSeconds := math.Pow(2, float64(count))
+
+			time.Sleep(time.Duration(sleepSeconds) * time.Second)
+
+			continue
+		}
+
+		break
+	}
+
+	return nil
 }
 
 func pushHeartbeats(filepath string, hh []heartbeat.Heartbeat) error {
--- a/pkg/offline/offline.go
+++ b/pkg/offline/offline.go
@@ -3,9 +3,11 @@ package offline
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
 	"github.com/wakatime/wakatime-cli/pkg/log"
@@ -15,8 +17,13 @@ import (
 )
 
 const (
+	// dbFilename is the default bolt db filename.
 	dbFilename = ".wakatime.bdb"
-	dbBucket   = "heartbeats"
+	// dbBucket is the standard bolt db bucket name.
+	dbBucket = "heartbeats"
+	// maxRequeueAttempts defines the maximum number of attempts to requeue heartbeats,
+	// which could not successfully be sent to the WakaTime API.
+	maxRequeueAttempts = 3
 	// sendLimit is the maximum number of heartbeats, which will be sent at once
 	// to the WakaTime API.
 	sendLimit = 24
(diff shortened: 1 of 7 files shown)


pkg/offline/offline.go
```diff
@@ -157,6 +157,146 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+		err = pushHeartbeats(filepath, hh[start:])
+		if err != nil {
+			log.Errorf("failed to push leftover heatbeats to queue: %s", err)
+		}
+	}
+
+	return err
+}
+
+func popHeartbeats(filepath string, syncLimit int) ([]heartbeat.Heartbeat, error) {
```
alanhamlett, line 235: This is great!

pkg/offline/offline.go
```diff
@@ -157,6 +157,146 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to start db transaction: %s", err)
+	}
+
+	queue, err := NewQueue(tx)
+	if err != nil {
+		return fmt.Errorf("failed initialize new queue: %s", err)
+	}
+
+	err = queue.PushMany(hh)
```
alanhamlett, line 331: Can we have some retry here? I'm thinking if there are any errors due to another process having the global boltdb lock, we should retry so heartbeats aren't lost.

pkg/offline/offline.go
```diff
@@ -157,6 +157,146 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+			result.Status != http.StatusAccepted &&
+			result.Status != http.StatusBadRequest {
+			withInvalidStatus = append(withInvalidStatus, hh[n])
+		}
+	}
+
+	if len(withInvalidStatus) > 0 {
+		log.Debugf("pushing %d heartbeat(s) with invalid result to queue", len(withInvalidStatus))
+
+		err = pushHeartbeats(filepath, withInvalidStatus)
```
alanhamlett, line 223: Maybe the retry goes here instead of in `pushHeartbeats`, up to you.

alanhamlett (approved):

pkg/offline/offline.go
alanhamlett: `syncLimit` should never be larger than `24`, but it looks like this comes directly from user input?

pkg/offline/offline.go
alanhamlett: `syncLimit` should just be `limit` because this code has nothing to do with syncing.. it's just popping multiple (up to 24) heartbeats from the offline db.

pkg/offline/offline.go
```diff
@@ -157,6 +157,146 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
 	}, nil
 }
 
+// Sync returns a function to send queued heartbeats to the WakaTime API.
+func Sync(filepath string, syncLimit int) func(next heartbeat.Handle) error {
```
alanhamlett, line 161: `syncLimit` should just be `limit` because this code has nothing to do with syncing.. it's just popping multiple (up to 24) heartbeats from the offline db.

pkg/offline/offline.go
alanhamlett: `syncLimit` should just be `limit` because this code has nothing to do with syncing.. it's just popping multiple (up to 24) heartbeats from the offline db.

dron22: Alright, after reading https://github.com/wakatime/wakatime-cli/pull/455#discussion_r650560600, I see the issue. I will add a constant `sendLimit` with value `24` here. Then I will add correct application of sync limit in https://github.com/wakatime/wakatime-cli/pull/455.

So what should be the default sync limit, in case `--sync-offline-activity` parameter is not provided?

pkg/offline/offline.go
dron22: Added a retry. Check new commits :+1:

pkg/offline/offline.go
dron22: Added a retry. Check new commits :+1:

pkg/offline/offline.go
alanhamlett: Default should be 1k.

pkg/offline/offline.go
```diff
@@ -157,6 +167,164 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+
+			for {
+				if count >= maxRequeueAttempts {
+					log.Errorf("abort requeuing after %d unsuccessful attempts", count)
+					break
+				}
+
+				requeueErr := pushHeartbeats(filepath, hh)
+				if requeueErr != nil {
+					log.Errorf("failed to push heatbeats after api error to queue: %s", requeueErr)
```
alanhamlett: This should be a debug log.

pkg/offline/offline.go
```diff
@@ -157,6 +167,164 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+			return nil
+		}
+
+		results, err := next(hh)
+		if err != nil {
+			var count int
+
+			for {
+				if count >= maxRequeueAttempts {
+					log.Errorf("abort requeuing after %d unsuccessful attempts", count)
```
alanhamlett: Let's change this to something like:

```go
log.Errorf("failed writing heartbeat to offline queue: %s", string(json.Marshal(hh)[0]))
```

That code probably doesn't work, but shows what we want is to log the heartbeat to `~/.wakatime.log` when unable to save to the offline db. That allows parsing the log file to recover lost heartbeats.

pkg/offline/offline.go
dron22: Alright, will adjust this in https://github.com/wakatime/wakatime-cli/pull/455

pkg/offline/offline.go
dron22: I guess we're best here with `warning`. I mean something really went wrong here, and also if we want to recover lost heartbeats, we don't want to depend on the user setting log level to debug, right?

alanhamlett: We've already got the error logging at `if count >= maxRequeueAttempts`, so we don't need to log two errors for the same thing. Here we can still retry, so the heartbeat might not be lost. If we retry, nothing should be printed unless `--verbose` was passed. That's because it's highly possible two wakatime-cli processes try to write at the same time, with one not able to acquire the global lock. That's not a problem, as long as we don't reach the `maxRequeueAttempts`.

pkg/offline/offline.go
```diff
@@ -157,6 +167,169 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+		err = pushHeartbeats(filepath, withInvalidStatus)
+		if err != nil {
+			log.Errorf("failed to push heatbeats with invalid status to queue: %s", err)
+		}
+	}
+
+	// handle leftover heartbeats
+	leftovers := len(hh) - len(results)
+	if leftovers > 0 {
+		log.Warnf("Missing %d results from api.", leftovers)
```
gandarez, line 232:
```suggestion
		log.Warnf("missing %d results from api", leftovers)
```

gandarez (commented):
one small remark

pkg/offline/offline.go
dron22: sharp eyes! :+1:

dron22: adjusted. see new commit below.

gandarez: Don't know why linter didn't catch it

gandarez (approved):

pkg/offline/offline.go
dron22: Yeah, this is a good idea. Adjusted following your remarks :+1:

PR #454, opened 2021-06-11
Remove offline sync from heartbeats pipeline (send all queued preparation pt2)

dron22:
This PR removes offline sync from heartbeats pipeline. Instead offline sync is performed after successful execution of heartbeats command. It is another preparation to sync all queued heartbeats on wakatime cli execution.

@alanhamlett Should we also perform offline sync, right after today/today-goal commands?

Has to be merged, only after merge of https://github.com/wakatime/wakatime-cli/pull/453

Related to: https://github.com/wakatime/wakatime-cli/issues/449

### DIFF
--- a/cmd/legacy/run.go
+++ b/cmd/legacy/run.go
@@ -105,7 +105,7 @@ func Run(cmd *cobra.Command, v *viper.Viper) {
 	if v.IsSet("entity") {
 		log.Debugln("command: heartbeat")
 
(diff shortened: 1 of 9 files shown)


alanhamlett:
> Should we also perform offline sync, right after today/today-goal commands?

No, we want those commands to return their stdout as soon as possible without waiting to sync offline heartbeats.

alanhamlett (approved):

gandarez (approved):

PR #455, opened 2021-06-12
Send all queued heartbeats in offline sync

### LINKED ISSUE #449: Send all queued heartbeats at once

dron22:
This PR changes offline sync to send all queued heartbeats to the WakaTime API.

- Any API results which are not 201/202/400 will be retried.
- Send limit of `24` is used as the maximum amount of heartbeats send per request
- Overall sync limit restricts the maximum number of heartbeats that will be sent in one wakatime-cli execution. It is set to a default of 1000 and can be provided via `--sync-offline-activity` flag. 

Closes #449

### DIFF
--- a/pkg/offline/offline.go
+++ b/pkg/offline/offline.go
@@ -17,6 +17,12 @@ import (
 )
 
 const (
+	// SyncMaxDefault is the default maximum number of heartbeats from the
+	// offline queue, which will be synced upon sending heartbeats to the API.
+	SyncMaxDefault = 1000
+)
+
+const (
 	// dbFilename is the default bolt db filename.
 	dbFilename = ".wakatime.bdb"
 	// dbBucket is the standard bolt db bucket name.
@@ -68,6 +74,8 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
 
 			results, err := next(hh)
 			if err != nil {
+				log.Debugf("pushing %d heartbeat(s) to queue due to error", len(hh))
+
 				requeueErr := pushHeartbeatsWithRetry(filepath, hh)
 				if requeueErr != nil {
 					log.Errorf("failed to push heatbeats to queue after api error: %s", requeueErr)
@@ -91,28 +99,55 @@ func Sync(filepath string, syncLimit int) func(next heartbeat.Handle) error {
 	return func(next heartbeat.Handle) error {
 		log.Debugf("execute offline sync with file %s", filepath)
 
-		hh, err := popHeartbeats(filepath, sendLimit)
-		if err != nil {
-			return fmt.Errorf("failed to fetch heartbeat from offline queue: %s", err)
-		}
+		var (
+			alreadySent int
+			run         int
+		)
 
-		if len(hh) == 0 {
-			log.Debugln("abort execution, as there are no queued heartbeats ready for sending")
+		for {
+			run++
 
-			return nil
-		}
-
-		results, err := next(hh)
-		if err != nil {
-			requeueErr := pushHeartbeatsWithRetry(filepath, hh)
-			if requeueErr != nil {
-				log.Warnf("failed to push heatbeats to queue after api error: %s", requeueErr)
+			if alreadySent >= syncLimit {
+				break
 			}
 
-			return err
+			var num = sendLimit
+
+			if alreadySent+sendLimit > syncLimit {
+				num = syncLimit - alreadySent
+				alreadySent += num
+			}
+
+			hh, err := popHeartbeats(filepath, num)
+			if err != nil {
+				return fmt.Errorf("failed to fetch heartbeat from offline queue: %s", err)
+			}
+
+			if len(hh) == 0 {
+				log.Debugln("no queued heartbeats ready for sending")
+
+				break
+			}
+
+			log.Debugf("send %d heartbeats on sync run %d", len(hh), run)
+
+			results, err := next(hh)
+			if err != nil {
+				requeueErr := pushHeartbeatsWithRetry(filepath, hh)
+				if requeueErr != nil {
+					log.Warnf("failed to push heatbeats to queue after api error: %s", requeueErr)
+				}
+
+				return err
+			}
+
+			err = handleResults(filepath, results, hh)
+			if err != nil {
+				return fmt.Errorf("failed to handle heatbeats api results: %s", err)
+			}
 		}
 
-		return handleResults(filepath, results, hh)
+		return nil
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/offline/offline.go
```diff
@@ -77,40 +78,81 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+		var (
+			count   int
+			exclude []string
+		)
 
-		if len(hh) == 0 {
-			log.Debugln("abort execution, as there are no queued heartbeats ready for sending")
+		for {
+			if count > maxSyncRequests {
+				return fmt.Errorf("exceeded max sync request limit of %d", maxSyncRequests)
```
alanhamlett: This should be warn level, not error.

pkg/offline/offline.go
```diff
@@ -217,6 +271,24 @@ func pushHeartbeats(filepath string, hh []heartbeat.Heartbeat) error {
 	return nil
 }
 
+// PopOptions contain options for poppoing heartbeats from the offline queue.
+type PopOptions struct {
+	Limit    int
+	Excluded []string
+}
+
+// Exclude returns true, if heartbeat with passed in id should be excluded.
```
alanhamlett: We don't need to check exclude again when sending offline heartbeats since it's already checked when creating the heartbeat. We also want to respect the `wakatime.cfg` state at time of the heartbeat's timestamp, where maybe yesterday the user wanted to include a project but today they don't. In that case, the offline heartbeats from yesterday should still be sent.

pkg/offline/offline.go
```diff
@@ -77,40 +78,81 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+			var ignored int
+
+			for _, i := range invalidResults {
+				// ignore "too many heartbeats" error results here, to send them again.
+				if i.Status == http.StatusTooManyRequests {
+					ignored++
+					continue
+				}
+
+				exclude = append(exclude, i.Heartbeat.ID())
```
alanhamlett: We only want to exclude heartbeats on `400` response code. Every other response code (`429`, `500`, etc.) we want to add the heartbeats back to the offline db.

pkg/offline/offline.go
```diff
@@ -260,6 +332,10 @@ func (q *Queue) PopMany(limit int) ([]heartbeat.Heartbeat, error) {
 			return nil, fmt.Errorf("failed to json unmarshal heartbeat data: %s", err)
 		}
 
+		if opts.Exclude(h.ID()) {
```
alanhamlett: We don't want to check exclude when sending offline heartbeats.

pkg/offline/offline.go
```diff
@@ -77,40 +78,81 @@ func WithQueue(filepath string, syncLimit int) (heartbeat.HandleOption, error) {
...
+		for {
+			if count > maxSyncRequests {
+				return fmt.Errorf("exceeded max sync request limit of %d", maxSyncRequests)
+			}
 
-			return nil
-		}
+			opts := PopOptions{
+				Excluded: exclude,
+				Limit:    syncLimit,
```
alanhamlett: Does `syncLimit` come from `--sync-offline-activity syncLimit`? That param should control the total heartbeats sent, but not how many we send per request. Per request, we should always send `<= 24` heartbeats.

pkg/offline/offline.go
alanhamlett: Instead of hard-coding `maxSyncRequests` let's use `maxSyncRequests := limit / 24 * 2`.

We also need to inspect the response code of each heartbeat in the bulk response. If all heartbeat responses in the bulk response are error codes, or the bulk response itself has an error code, then we should break and stop sending offline heartbeats.

dron22: > Instead of hard-coding maxSyncRequests let's use maxSyncRequests := limit / 24 * 2.

So this would restrict it to 2 requests. But if the user has more than 48 heartbeats in the queue, we want to sync them all, right?

pkg/offline/offline.go
dron22: Ah alright, I didn't know this. Valuable info! Will adjust :+1:

pkg/offline/offline.go
dron22: > We also need to inspect the response code of each heartbeat in the bulk response. If all heartbeat responses in the bulk response are error codes, or the bulk response itself has an error code, then we should break and stop sending offline heartbeats.

Alright, to sum it up and please correct me if I am missing something:
1. Response global error: push heartbeats back to queue and abort sync
2. All results with invalid status: push heartbeats back to queue and abort sync
3. 429/500 result: push heartbeat back to queue and continue sync, but do not send this heartbeat again in this cli execution.
4. 400: error log, reject heartbeat and continue sync

pkg/offline/offline.go
dron22: Maybe using the term exclude here is a little misleading. But here I keep track of all heartbeats, that were already sent during this cli execution.

In this implementation, as a safety measure, every heartbeat will only be sent once in one wakatime-cli execution. Every heartbeat, which was pushed back to the queue will not be sent again in the same wakatime-cli execution.

Reason: As pop and push are separated, we might end up in an endless loop trying to sync the same heartbeats over and over again.

pkg/offline/offline.go
alanhamlett: > 3. 429/500 result: push heartbeat back to queue and continue sync, but do not send this heartbeat again in this cli execution.

The error code doesn't matter, as long as it's not 400. On any error result, put the heartbeat at the back of the queue and continue syncing until the total heartbeats sent reaches `--sync-offline-activity limit`. Each heartbeat sent (failure or success) increases the total count approaching `--sync-offline-activity limit`. If we haven't reached `limit` yet, we can attempt to re-send non-400 heartbeats again in the same cli execution.

pkg/offline/offline.go
alanhamlett: We don't need to track heartbeats already sent, we just put them at the back of the queue and allow re-sending them or don't put them back into the queue if a heartbeat returned 400.

pkg/offline/offline.go
alanhamlett: `--sync-offline-activity limit` is doing the same as your `As a safety measure, to prevent an endless loop, a maximum sync request limit of 100 was added`, so we don't really need the 100 limit.

pkg/offline/offline.go
dron22: :+1:

alanhamlett (approved):

PR #465, opened 2021-06-25
New --offline-count argument

### LINKED ISSUE #462: New command line argument `--offline-count` to print number of heartbeats currently in offline db
For ex:

```
$ wakatime-cli --offline-count
94
```

alanhamlett:
Prints the number of heartbeats currently in the offline db, then exits.

Closes #462.

### DIFF
--- a/cmd/legacy/run.go
+++ b/cmd/legacy/run.go
@@ -14,6 +14,7 @@ import (
 	"github.com/wakatime/wakatime-cli/cmd/legacy/legacyapi"
 	"github.com/wakatime/wakatime-cli/cmd/legacy/legacyparams"
 	"github.com/wakatime/wakatime-cli/cmd/legacy/logfile"
+	"github.com/wakatime/wakatime-cli/cmd/legacy/offlinecount"
 	"github.com/wakatime/wakatime-cli/cmd/legacy/offlinesync"
 	"github.com/wakatime/wakatime-cli/cmd/legacy/today"
 	"github.com/wakatime/wakatime-cli/cmd/legacy/todaygoal"
@@ -114,10 +115,17 @@ func Run(cmd *cobra.Command, v *viper.Viper) {
 		RunCmd(v, offlinesync.Run)
 	}
 
+	if v.GetBool("offline-count") {
+		log.Debugln("command: offline-count")
+
+		RunCmd(v, offlinecount.Run)
+	}
+
 	log.Warnf("One of the following parameters has to be provided: %s", strings.Join([]string{
 		"--config-read",
 		"--config-write",
 		"--entity",
+		"--offline-count",
 		"--sync-offline-activity",
 		"--today",
 		"--today-goal",
--- a/main_test.go
+++ b/main_test.go
@@ -225,6 +225,68 @@ func TestTodaySummary(t *testing.T) {
 	assert.Eventually(t, func() bool { return numCalls == 1 }, time.Second, 50*time.Millisecond)
 }
 
+func TestOfflineCountEmpty(t *testing.T) {
+	offlineQueueFile, err := ioutil.TempFile(os.TempDir(), "")
+	require.NoError(t, err)
+
+	defer os.Remove(offlineQueueFile.Name())
+
+	out := run(exec.Command(
+		binaryPath(t),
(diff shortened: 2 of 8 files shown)


cmd/legacy/run.go
```diff
@@ -114,11 +115,18 @@ func Run(cmd *cobra.Command, v *viper.Viper) {
...
+
+		RunCmd(v, offlinecount.Run)
+	}
+
 	log.Warnf("One of the following parameters has to be provided: %s", strings.Join([]string{
 		"--config-read",
 		"--config-write",
 		"--entity",
 		"--sync-offline-activity",
+		"--offline-count",
```
dron22: Just to maintain the alphabetic sorting.
```suggestion
		"--offline-count",
		"--sync-offline-activity",
```

main_test.go
```diff
@@ -225,6 +225,17 @@ func TestTodaySummary(t *testing.T) {
 	assert.Eventually(t, func() bool { return numCalls == 1 }, time.Second, 50*time.Millisecond)
 }
 
+func TestOfflineCount(t *testing.T) {
```
dron22: Here it would be good to use a temporary file as offline queue like in `testSendHeartbeats` (https://github.com/wakatime/wakatime-cli/blob/develop/main_test.go#L109). Otherwise, when running it locally, it will use the default file, which might have heartbeats inside.

pkg/offline/offline.go
```diff
@@ -322,11 +345,11 @@ type Queue struct {
 }
 
 // NewQueue creates a new instance of Queue.
-func NewQueue(tx *bolt.Tx) (*Queue, error) {
+func NewQueue(tx *bolt.Tx) *Queue {
 	return &Queue{
 		Bucket: dbBucket,
 		tx:     tx,
-	}, nil
+	}
```
dron22, line 356: :+1:

pkg/offline/offline_test.go
```diff
@@ -950,6 +944,25 @@ func TestQueue_PushMany(t *testing.T) {
...
+
+	tx, err := db.Begin(true)
+	require.NoError(t, err)
+
+	q := offline.NewQueue(tx)
+	q.Bucket = "test_bucket"
+
+	count, err := q.Count()
+	require.NoError(t, err)
+	assert.Equal(t, count, 0)
```
dron22: Here it would be interesting to test it against an offline queue with heartbeats inside.

main_test.go
dron22: Also testing it against an offline queue with heartbeats inside, would be a little more meaningful.

dron22 (commented):
Nice one :+1: Looks clean. Just added some comments regarding tests.

pkg/offline/offline.go
```diff
@@ -313,6 +307,35 @@ func pushHeartbeats(filepath string, hh []heartbeat.Heartbeat) error {
...
+
+	defer db.Close()
+
+	tx, err := db.Begin(true)
+	if err != nil {
+		return 0, fmt.Errorf("failed to start db transaction: %s", err)
+	}
+
+	queue := NewQueue(tx)
+	count, err := queue.Count()
```
gandarez, line 326: `err` here could be checked right after calling `queue.Count()`.

```Go
if err != nil {
    log.Errorf("failed to count offline queue: %s", err)
}
```

pkg/offline/offline.go
```diff
@@ -390,3 +413,14 @@ func (q *Queue) PushMany(hh []heartbeat.Heartbeat) error {
 
 	return nil
 }
+
+// Count returns the total number of heartbeats in the offline db.
+func (q *Queue) Count() (int, error) {
+	b, err := q.tx.CreateBucketIfNotExists([]byte(q.Bucket))
+	if err != nil {
+		log.Errorf("failed to create/load bucket: %s", err)
+		return 0, fmt.Errorf("failed to create/load bucket: %s", err)
```
gandarez, line 425: Instead of duplicating error message here it could only return and the error would be checked on the caller.

gandarez (commented):
just one small remark to make error handling cleaner.

gandarez (approved):

PR #467, opened 2021-06-25
Prevent writing to default log file in integration tests

dron22:
This PR adjusts integration test code to write logs to a temporary file, instead of the default log file.


dron22:
Replaced by https://github.com/wakatime/wakatime-cli/pull/468 due to invalid branch name

PR #468, opened 2021-06-25
Prevent writing to default log file in integration tests

dron22:
This PR adjusts integration test code to write logs to a temporary file, instead of the default log file.


main_test.go
```diff
@@ -253,7 +249,26 @@ func TestVersionVerbose(t *testing.T) {
 	)), out)
 }
 
-func run(cmd *exec.Cmd) string {
+func runWakatime(t *testing.T, args ...string) string {
```
gandarez: Maybe `runWakatime` sounds too genereic. What about `runWakatimeCli` or `runBinary`?

gandarez (approved):
Already approving but just left a small remark.

PR #469, opened 2021-06-27
Log warning on system cert pool error

### LINKED ISSUE #46937: crypto/x509: unable to verify proxy self-signed cert
### What version of Go are you using (`go version`)?

<pre>
$ go version
1.16.5
</pre>

### Does this issue reproduce with the latest release?
Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GOARCH="amd64"
GOOS="windows"
</pre></details>

I personally don't use Windows, this issue occurs on users' Windows machines. It most likely also occurs on darwin, linux, etc. too.

### What did you do?

We first learned of this with https://github.com/wakatime/wakatime-cli/issues/385. The problem is Go's [crypto/tls](https://golang.org/pkg/crypto/tls/) doesn't select the black-box corporate proxy's cert with the error:
`x509: certificate signed by unknown authority`

If we disable SSL verification, the request goes through. Python and Git [can connect just fine](https://github.com/wakatime/wakatime-cli/issues/385#issuecomment-856951861) however Curl has some issues too but curl works with the `--ssl-no-revoke` flag.

My hypotheses:

* Go thinks the Proxy's cert (installed correctly on the user's machine by their IT) is not valid for the server's domain. Maybe because the Proxy server doesn't rewrite the Host header.
* The Proxy's cert is valid for domains `*`, but Go needs a TLD?
* Some other reason Go thinks the Proxy's cert isn't valid for the end destination that's different from Python/OpenSSL. The error message means a cert for the domain wasn't found, not that one was found invalid.

### What did you expect to see?
A valid TLS connection through the proxy.

### What did you see instead?
Error message `x509: certificate signed by unknown authority`

alanhamlett:
Anytime unable to use system cert pool, log a warning to `~/.wakatime.log` to fix https://github.com/golang/go/issues/46937#issuecomment-869066358.

### DIFF
--- a/pkg/api/transport.go
+++ b/pkg/api/transport.go
@@ -4,6 +4,8 @@ import (
 	"crypto/x509"
 	"net/http"
 	"time"
+
+	"github.com/wakatime/wakatime-cli/pkg/log"
 )
 
 // NewTransport initializes a new http.Transport.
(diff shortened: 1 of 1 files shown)


gandarez (approved):
LGTM

PR #475, opened 2021-07-02
Missing files are warnings not errors

alanhamlett:
This prevents logging an error when ignoring a file, and prevents sending diagnostics.


cmd/legacy/run.go
```diff
@@ -138,6 +119,30 @@ func Run(cmd *cobra.Command, v *viper.Viper) {
 	os.Exit(exitcode.ErrDefault)
 }
 
+// SetupLogging uses the --log-file param to configure logging to file or stdout.
+func SetupLogging(v *viper.Viper) {
```
gandarez, line 123: Maybe we'll need some test cases for this func in the future. It's a critical part of the app.

gandarez (approved):

PR #479, opened 2021-07-02
Read INI config file before setting up logging

alanhamlett:
This fixes a bug where `debug = true` wasn't enabling debug logs because logging was setup before reading the config file.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,10 @@ linters:
     - unparam
     - whitespace
     - wsl
+linters-settings:
+  gosec:
+    excludes:
+      - G204
 issues:
   include:
     - EXC0002
--- a/cmd/legacy/run.go
+++ b/cmd/legacy/run.go
@@ -32,6 +32,12 @@ import (
 
 // Run executes legacy commands following the interface of the old python implementation of the WakaTime script.
 func Run(cmd *cobra.Command, v *viper.Viper) {
+	if err := config.ReadInConfig(v, config.FilePath); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to load configuration file: %s", err)
+
+		os.Exit(exitcode.ErrConfigFileParse)
+	}
+
 	SetupLogging(v)
 
 	if v.GetBool("useragent") {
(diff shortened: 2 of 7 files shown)


cmd/legacy/run.go
```diff
@@ -123,6 +123,7 @@ func Run(cmd *cobra.Command, v *viper.Viper) {
 func SetupLogging(v *viper.Viper) {
 	logfileParams, err := logfile.LoadParams(v)
 	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to load log params: %s", err)
```
gandarez, line 126: Doesn't `log.*` package write to console (output) when not already setup?

.golangci.yml
```diff
@@ -19,6 +19,10 @@ linters:
     - unparam
     - whitespace
     - wsl
+linters-settings:
+  gosec:
+    excludes:
+      - G204
```
gandarez, line 25: Why do we need to exclude G204? What has changed?

main_test.go
```diff
@@ -131,12 +129,13 @@ func testSendHeartbeats(t *testing.T, entity, project string) {
...
 
 func TestTodayGoal(t *testing.T) {
-	apiUrl, router, close := setupTestServer()
+	apiURL, router, close := setupTestServer()
 	defer close()
 
 	var numCalls int
 
-	router.HandleFunc("/users/current/goals/11111111-1111-4111-8111-111111111111", func(w http.ResponseWriter, req *http.Request) {
+	url := "/users/current/goals/11111111-1111-4111-8111-111111111111"
```
gandarez: I don't see that variable being used in other places. Do we need to have the url declared as variable?

alanhamlett:
Here's the actual fix: https://github.com/wakatime/wakatime-cli/pull/479/files#diff-c8d024d333624f9b55a13c37ddd2692852f5e2f1a6390dffb3c0af46f0dd75f9R35

Everything else is just linter, tests, and minor logging/stdout fixes.

gandarez (commented):
Left some comments.

.golangci.yml
alanhamlett: that's for `main_test.go` which we aren't linting currently but hopefully we can fix that in the future.

main_test.go
alanhamlett: It's to prevent the line too long linter error

cmd/legacy/run.go
alanhamlett: Oh maybe, but I like to be explicit. Also I wasn't sure if it would write to stdout or stderr, and we want errors always written to stderr.

PR #519, opened 2021-08-28
Add backoff algorithm

### LINKED ISSUE #510: Implement Exponential Backoff
When seding a hertbeat (normal and offline) it first check `backoff_seconds` and `backoff_at` from `.wakatime.cfg` at `internal` section. If it still in backoff, save heartbeat(s) to offline db and exit. Otherwise, send heartbeats as normal.

The formula:
```code
(now - backoff_at >= backoff_seconds) OR (now - backoff_at > 3600)
```

Screen Shot 2021-08-18 at 3 58 32 PM

gandarez:
This PR adds exponential backoff algorithm to prevent trying to send heartbeats during backoff period. It will act in the pipeline after offline step and before calling the API. If any error returns from the API, the backoff step will decide to either increment retry or reset it after `reset time` of 3600 seconds. If no error returns from the API and in backoff then reset its settings.

Default values used:
* **3600** seconds is the maximum time backoff will last for.
* **15** is the factor to multiply by retries.

Closes #510

### DIFF
--- a/cmd/legacy/legacyparams/params.go
+++ b/cmd/legacy/legacyparams/params.go
@@ -9,10 +9,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/mitchellh/go-homedir"
 	"github.com/wakatime/wakatime-cli/pkg/api"
+	"github.com/wakatime/wakatime-cli/pkg/config"
+	"github.com/wakatime/wakatime-cli/pkg/log"
 	"github.com/wakatime/wakatime-cli/pkg/vipertools"
 
+	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/viper"
 )
 
@@ -40,6 +42,8 @@ type Params struct {
 
 // API contains api related parameters.
 type API struct {
+	BackoffAt        time.Time
+	BackoffRetries   int
 	DisableSSLVerify bool
 	Hostname         string
 	Key              string
@@ -52,11 +56,19 @@ type API struct {
 
 // String implements fmt.Stringer interface.
 func (p API) String() string {
+	var backoffAt string
+	if !p.BackoffAt.IsZero() {
+		backoffAt = p.BackoffAt.Format(config.DateFormat)
+	}
+
 	return fmt.Sprintf(
-		"api key: '%s', api url: '%s', hostname: '%s', plugin: '%s', timeout: %s,"+
-			" disable ssl verify: %t, proxy url: '%s', ssl cert filepath: '%s'",
+		"api key: '%s', api url: '%s', backoff at: '%s', backoff retries: %d,"+
+			" hostname: '%s', plugin: '%s', timeout: %s, disable ssl verify: %t,"+
+			" proxy url: '%s', ssl cert filepath: '%s'",
 		p.Key[:4]+"...",
 		p.URL,
+		backoffAt,
+		p.BackoffRetries,
 		p.Hostname,
 		p.Plugin,
 		p.Timeout,
@@ -133,6 +145,20 @@ func loadAPIParams(v *viper.Viper) (API, error) {
 	apiURL = strings.TrimSuffix(apiURL, "/heartbeats")
 	apiURL = strings.TrimSuffix(apiURL, "/heartbeat")
 
+	var backoffAt time.Time
+
+	backoffAtStr := vipertools.GetString(v, "internal.backoff_at")
+	if backoffAtStr != "" {
+		parsed, err := time.Parse(config.DateFormat, backoffAtStr)
+		if err != nil {
+			log.Warnf("failed to parse backoff_at: %s", err)
+		} else {
+			backoffAt = parsed
+		}
+	}
+
+	backoffRetries, _ := vipertools.FirstNonEmptyInt(v, "internal.backoff_retries")
+
 	var (
 		hostname string
 		err      error
@@ -178,6 +204,8 @@ func loadAPIParams(v *viper.Viper) (API, error) {
 	}
 
 	return API{
+		BackoffAt:        backoffAt,
+		BackoffRetries:   backoffRetries,
 		DisableSSLVerify: vipertools.FirstNonEmptyBool(v, "no-ssl-verify", "settings.no_ssl_verify"),
 		Hostname:         hostname,
 		Key:              apiKey,
--- a/cmd/legacy/legacyparams/params_test.go
+++ b/cmd/legacy/legacyparams/params_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/wakatime/wakatime-cli/cmd/legacy/legacyparams"
 	"github.com/wakatime/wakatime-cli/pkg/api"
+	"github.com/wakatime/wakatime-cli/pkg/config"
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
@@ -159,6 +160,7 @@ func TestLoad_API_APIKey(t *testing.T) {
 			ViperAPIKeyConfigOld: "20000000-0000-4000-8000-000000000000",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "https://api.wakatime.com/api/v1",
 					Hostname: "my-computer",
@@ -170,6 +172,7 @@ func TestLoad_API_APIKey(t *testing.T) {
 			ViperAPIKeyConfigOld: "10000000-0000-4000-8000-000000000000",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "https://api.wakatime.com/api/v1",
 					Hostname: "my-computer",
@@ -180,6 +183,7 @@ func TestLoad_API_APIKey(t *testing.T) {
 			ViperAPIKeyConfigOld: "00000000-0000-4000-8000-000000000000",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "https://api.wakatime.com/api/v1",
 					Hostname: "my-computer",
@@ -241,6 +245,7 @@ func TestLoad_API_APIUrl(t *testing.T) {
 			ViperAPIUrlOld:    "http://localhost:8082",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "http://localhost:8080",
 					Hostname: "my-computer",
@@ -252,6 +257,7 @@ func TestLoad_API_APIUrl(t *testing.T) {
 			ViperAPIUrlOld:    "http://localhost:8082",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "http://localhost:8082",
 					Hostname: "my-computer",
@@ -262,6 +268,7 @@ func TestLoad_API_APIUrl(t *testing.T) {
 			ViperAPIUrlConfig: "http://localhost:8081",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "http://localhost:8081",
 					Hostname: "my-computer",
@@ -272,6 +279,7 @@ func TestLoad_API_APIUrl(t *testing.T) {
 			ViperAPIUrl: "http://localhost:8080/api/v1/heartbeats.bulk",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "http://localhost:8080/api/v1",
 					Hostname: "my-computer",
@@ -282,6 +290,7 @@ func TestLoad_API_APIUrl(t *testing.T) {
 			ViperAPIUrl: "http://localhost:8080/api/",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "http://localhost:8080/api",
 					Hostname: "my-computer",
@@ -292,6 +301,7 @@ func TestLoad_API_APIUrl(t *testing.T) {
 			ViperAPIUrl: "http://localhost:8080/api/heartbeat",
 			Expected: legacyparams.Params{
 				API: legacyparams.API{
+
 					Key:      "00000000-0000-4000-8000-000000000000",
 					URL:      "http://localhost:8080/api",
 					Hostname: "my-computer",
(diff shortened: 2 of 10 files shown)


cmd/legacy/legacyparams/params.go
```diff
@@ -133,6 +145,20 @@ func loadAPIParams(v *viper.Viper) (API, error) {
...
+	var backoffAt *time.Time
+
+	backoffAtStr := vipertools.GetString(v, "internal.backoff_at")
+	if backoffAtStr != "" {
+		parsed, err := time.Parse(config.DateFormat, backoffAtStr)
+		if err != nil {
+			log.Warnf("failed to parse backoff at to time instance: %s", err)
+		}
+
+		backoffAt = &parsed
```
dron22: I guess in case of failing to parse the string, it is better to ignore the returned value. Although, in go if/else is mostly avoided, I don't think, we can get around it here. 
```suggestion
		if err != nil {
			log.Warnf("failed to parse backoff at to time instance: %s", err)
		} else {
		    backoffAt = &parsed
		}
```

draft/structure.md
```diff
@@ -1,107 +1,124 @@
...
-                func detectWithPlugin(filepath string, plugin Plugin) (project, branch string, err error)
-        summary/
-            summary.go
-                type Summary struct {}
-        version/
-            version.go
-    go.mod
-    go.sum
-    main.go
+wakatime-cli/
```
dron22: I guess we can delete `/draft` folder, as this was only needed for planning the first version.

cmd/legacy/legacyparams/params_test.go
```diff
@@ -330,6 +331,45 @@ func TestLoad_APIUrl_Default(t *testing.T) {
...
+	v := viper.New()
+	v.SetDefault("sync-offline-activity", 1000)
+	v.Set("key", "00000000-0000-4000-8000-000000000000")
+	v.Set("hostname", "my-computer")
+	v.Set("internal.backoff_at", "2021-08-27T18:50:42-03:00")
+
+	params, err := legacyparams.Load(v)
+	require.NoError(t, err)
+
+	backoffAt, err := time.Parse(config.DateFormat, "2021-08-27T18:50:42-03:00")
```
dron22: Oh whao, I didn't know, that `assert.Equal` works with pointers in this way :+1:

main_test.go
```diff
@@ -109,11 +109,16 @@ func testSendHeartbeats(t *testing.T, entity, project string) {
...
+	require.NoError(t, err)
+
+	defer os.Remove(tmpFile.Name())
+
 	runWakatimeCli(
 		t,
 		"--api-url", apiURL,
 		"--key", "00000000-0000-4000-8000-000000000000",
-		"--config", "testdata/wakatime.cfg",
+		"--config", tmpFile.Name(),
```
dron22, line 120: :+1:

pkg/backoff/backoff_internal_test.go
```diff
@@ -0,0 +1,85 @@
...
+	tmpFile, err := ioutil.TempFile(os.TempDir(), "wakatime")
+	require.NoError(t, err)
+
+	defer os.Remove(tmpFile.Name())
+
+	v.Set("config", tmpFile.Name())
+
+	at := time.Now().Add(time.Minute*(resetAfter*-1) + time.Second*-1)
+
+	err = updateBackoffSettings(v, 2, &at)
```
dron22: What is the advantage of passing `*time.Time` instead of `time.Time` into `updateBackoffSettings`?

pkg/backoff/backoff_internal_test.go
```diff
@@ -0,0 +1,85 @@
...
+	err = updateBackoffSettings(v, 2, nil)
+	require.NoError(t, err)
+
+	ini, err := config.NewIniWriter(v, func(vp *viper.Viper) (string, error) {
+		assert.Equal(t, v, vp)
+		return tmpFile.Name(), nil
+	})
+	require.NoError(t, err)
+
+	assert.NotEmpty(t, ini.File.Section("internal").Key("backoff_at").String())
```
dron22: Let's assert the actual value with `assert.WithinDuration` (https://pkg.go.dev/github.com/stretchr/testify/assert#WithinDuration). As you cannot predict the now time, `WithinDuration` will allow you to ensure that the new config value is a certain timeframe (e.g. now +/- one second)

pkg/backoff/backoff_internal_test.go
```diff
@@ -0,0 +1,85 @@
...
+	err = updateBackoffSettings(v, 2, &at)
+	require.NoError(t, err)
+
+	ini, err := config.NewIniWriter(v, func(vp *viper.Viper) (string, error) {
+		assert.Equal(t, v, vp)
+		return tmpFile.Name(), nil
+	})
+	require.NoError(t, err)
+
+	assert.NotEmpty(t, ini.File.Section("internal").Key("backoff_at").String())
```
dron22: same here, see comment above on `assert.WithinDuration`

pkg/backoff/backoff_internal_test.go
```diff
@@ -0,0 +1,85 @@
...
+	err = resetBackoffSettings(v)
+	require.NoError(t, err)
+
+	ini, err := config.NewIniWriter(v, func(vp *viper.Viper) (string, error) {
+		assert.Equal(t, v, vp)
+		return tmpFile.Name(), nil
+	})
+	require.NoError(t, err)
+
+	assert.NotEmpty(t, ini.File.Section("internal").Key("backoff_at").String())
```
dron22: same here, see comment above on `assert.WithinDuration`

pkg/backoff/backoff_test.go
```diff
@@ -0,0 +1,87 @@
...
+	handle := opt(func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
+		return []heartbeat.Result{
+			{
+				Status: 201,
+			},
+		}, nil
+	})
+
+	_, err := handle([]heartbeat.Heartbeat{})
+	require.Error(t, err)
```
dron22, line 61: Here it could make sense to use a custom error like `backoff.ErrBackoffActive`. This would allow to differentiate this from a general error on sending by previous heartbeat pipeline steps. What do you think?

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,123 @@
...
+	// Retries is the number of attempts to connect.
+	Retries int
+	// V is an instance of Viper.
+	V *viper.Viper
+}
+
+// WithRetry initializes and returns a heartbeat handle option, which
+// can be used in a heartbeat processing pipeline to prevent trying to send
+// a heartbeat when the api is unresponsive.
+func WithRetry(config RetryConfig) heartbeat.HandleOption {
```
dron22: Great! Very good idea to include it as a pipeline handle option :+1:

dron22 (commented):
I like the general solution :+1:  Left comments

cmd/legacy/legacyparams/params.go
gandarez: you're right. I didn't pay attention when setting the parsed value after an error.

cmd/legacy/legacyparams/params_test.go
gandarez: me neither lol

pkg/backoff/backoff_internal_test.go
gandarez: The `backoff_at` key can be a null value when it's not present in the config file or parsing failed. Since it can be null, a pointer is necessary to the logic for backoff to work. The possible scenarios are:
* not in backoff, api returns success, then nothing changes.
* not in backoff, api returns an error, then it updates (adds) backoff settings. if BackoffAt is a null value it stamps a new time.
* in backoff, api returns an error, then it updates backoff settings. BackoffAt is **not** a null value.
* in backoff, api returns an error, limit time during backoff reached, then it resets backoff settings to retries 1 and stamp a new time. BackoffAt is **not** a null value.

pkg/backoff/backoff_internal_test.go
gandarez: Ohh didn't know about that function. Good to know, thanks.

pkg/backoff/backoff.go
gandarez: Learned with you :)

pkg/backoff/backoff_test.go
gandarez: Being honest I haven't decided to create a new custom error because even if an error during backoff occurs the offline step needs to enque the current heartbeat. If you have other thoughts please let me know.

dron22: yeah, you're right, from the perspective of offline queue, there is no difference actually. :+1:

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,123 @@
...
+	return func(next heartbeat.Handle) heartbeat.Handle {
+		return func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
+			log.Debugln("execute heartbeat backoff algorithm")
+
+			nextDuration, _ := time.ParseDuration(fmt.Sprintf("%fs", float64(factor)*math.Pow(2, float64(config.Retries))))
```
dron22: To avoid string creation/parsing, it can be simplified to:
```suggestion
			nextDuration, _ := time.Duration(float64(factor)*math.Pow(2, float64(config.Retries))) * time.Second
```

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,123 @@
...
+
+			now := time.Now()
+			if config.At != nil && now.Before(config.At.Add(nextDuration)) {
```
dron22: I guess we want to include `resetAfter` here, to ensure, that we send the heartbeats after the defined time. Otherwise we will not reset after 3600 seconds.
```suggestion
			if config.At != nil && now.Before(config.At.Add(nextDuration)) && now.Before(config.At.Add(resetAfter)) {
```

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func resetBackoffSettings(v *viper.Viper) error {
+	w, err := ini.NewIniWriter(v, ini.FilePath)
+	if err != nil {
+		return fmt.Errorf("failed to parse config file: %s", err)
+	}
+
+	keyValue := map[string]string{
+		"backoff_at":      time.Now().Format(ini.DateFormat),
```
dron22: We reset here. Reset ends the backoff, thus after resetting in `WithRetry`, `config.At` should be `nil` to not cause any backoff.

But here we set `backoff_at` to a timestamp. `config.At` won't be `nil` in `WithRetry` then, or am I missing something?

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,123 @@
...
+type RetryConfig struct {
+	// At is the time when the first failure happened.
+	At *time.Time
+	// Retries is the number of attempts to connect.
+	Retries int
+	// V is an instance of Viper.
+	V *viper.Viper
+}
+
+// WithRetry initializes and returns a heartbeat handle option, which
```
dron22: We prevent sending heartbeats here, so actually we're implementing a circuit breaker here (https://microservices.io/patterns/reliability/circuit-breaker.html).

It is not really about a retrying here. Retrying only happens due to offline pipeline step. But the backoff pipeline step should not depend on offline step being present or not. So how about changing the name to `WithBackoff` or something  similar? 

```suggestion
// WithBackoff initializes and returns a heartbeat handle option, which
```

dron22 (commented):
Alright, thanks for the adjustments. I went deeper into the backoff algorithm and added some additional comments.

pkg/backoff/backoff.go
gandarez: I believe we don't need to ensure that current time is before last backoff time because at `updateBackoffSettings` it resets backoff in case of current time is greater than the reset limit.

```Go
// reset backoff settings if reached max allowed time
if at != nil && now.Sub(*at).Seconds() > resetAfter {
	keyValue["backoff_at"] = now.Format(ini.DateFormat)
	keyValue["backoff_retries"] = "1"
}
```

pkg/backoff/backoff.go
gandarez: To get backoff reset means no error has returned from the API, thus the process will exit with zero which means no retry will be done at current process. Tne next execution will load again the params and go on..

pkg/backoff/backoff.go
gandarez: humm not too sure if it's a circuit breaker. Most of the articles refer backoff as a retry strategy and circuit breaker as a gate to return success or not in case of failures. Even if wakatime-cli runs once per request (instead of an API that is always on) it controls backoff by applying a retry policy persisted as `backoff_retries` and `backoff_at`. IMHO it's more a retry than a circuit breaker. Think about offline step, it also breaks the execution if fails.

pkg/backoff/backoff.go
alanhamlett: Shouldn't this reset to zero instead of `1`?

alanhamlett: Also, we should set `backoff_at ` to null when resetting to have no backoff on the next try.

gandarez: We never save `retries = 0`, so cleaning retries will make cli to consider it as zero value. @alanhamlett reset will be fixed to not save current time. Thanks for pointing out.

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,126 @@
...
+		"backoff_retries": strconv.Itoa(retries),
+	}
+
+	now := time.Now()
+
+	if at == nil {
+		keyValue["backoff_at"] = now.Format(ini.DateFormat)
+	}
+
+	// reset backoff settings if reached max allowed time
```
alanhamlett: Let's just remove this reset block, and only reset on successful API response && when retries is non-null and greater than zero.

pkg/backoff/backoff.go
dron22: But this particular step does not control the retry. The only thing it does is to block the call or not and increment a count.

alanhamlett (approved):

pkg/backoff/backoff.go
alanhamlett: renamed

alanhamlett (changes requested):
When testing locally I'm seeing this error:
```
failed to update backoff settings: failed to parse config file: error loading config file: key-value delimiter not found: COMMIT_EDITMSG$\n
```

My `~/.wakatime.cfg` file:
```
[settings]
ignore =
	COMMIT_EDITMSG$
	PULLREQ_EDITMSG$
	MERGE_MSG$
	TAG_EDITMSG$
debug = true
```

pkg/backoff/backoff.go
dron22: Latest changes fixed the problem :+1:

pkg/backoff/backoff.go
dron22: Latest changes fix the problem, which I pointed out :+1:

dron22:
@alanhamlett The changes of @gandarez, which fixed the logic and introduced `shouldBackoff` function disappeared in the branch. Was this intended?

gandarez:
After commits disappeared I reverted back to the original solution with adjustments discussed here.

gandarez:
> When testing locally I'm seeing this error:
> 
> ```
> failed to update backoff settings: failed to parse config file: error loading config file: key-value delimiter not found: COMMIT_EDITMSG$\n
> ```
> 
> My `~/.wakatime.cfg` file:
> 
> ```
> [settings]
> ignore =
> 	COMMIT_EDITMSG$
> 	PULLREQ_EDITMSG$
> 	MERGE_MSG$
> 	TAG_EDITMSG$
> debug = true
> ```

Should it be considered a new bug?

pkg/backoff/backoff.go
```diff
@@ -0,0 +1,124 @@
...
+	return func(next heartbeat.Handle) heartbeat.Handle {
+		return func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
+			log.Debugln("execute heartbeat backoff algorithm")
+
+			nextDuration := time.Duration(float64(factor)*math.Pow(2, float64(config.Retries))) * time.Second
+
+			now := time.Now()
+			if config.At != nil && now.Before(config.At.Add(nextDuration)) &&
+				now.Before(config.At.Add(resetAfter*time.Second)) {
+				log.Debugf("current time '%s' is before next backoff at '%s'",
```
dron22: I guess you mean `backoff end` or `retry` here, right?
```suggestion
				log.Debugf("current time '%s' is before next backoff end at '%s'",
```

dron22 (approved):
Whao, a lot of pushes going on here during review. At the current state, all changes look good for me :+1:

pkg/backoff/backoff.go
gandarez: Maybe only `"current time '%s' is before next backoff"`.

alanhamlett: I changed this in 89a70b9b0868ed65e9d6eaeb7c0b690233f6509d.

alanhamlett (approved):

PR #521, opened 2021-08-30
Increase default timeout to 120s

gandarez:
This PR increases default timeout to 120 seconds.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -4,6 +4,7 @@ import (
 	"strconv"
 
 	"github.com/wakatime/wakatime-cli/cmd/legacy"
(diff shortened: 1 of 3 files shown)


alanhamlett:
Let's also use `defaultTimeout` here:

* https://github.com/wakatime/wakatime-cli/blob/55b72ad417ad81634b319a4e80d4c84c592e7e94/pkg/api/transport.go#L15
* https://github.com/wakatime/wakatime-cli/blob/55a3c9da8beecff8a0c45fc0a5c5209e620e57ab/pkg/api/api.go#L37

alanhamlett (approved):

PR #523, opened 2021-08-31
Use new statusbar today endpoint

alanhamlett:
To reduce load on the API servers, let's use a dedicated API endpoint for status bar today stats so these requests can be cached separately from dashboard and other Summaries requests.


cmd/legacy/today/today.go
```diff
@@ -59,18 +58,14 @@ func Summary(v *viper.Viper) (string, error) {
 		return "", fmt.Errorf("failed to initialize api client: %w", err)
 	}
 
-	now := time.Now()
-	todayStart := time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, now.Location())
-	todayEnd := time.Date(now.Year(), now.Month(), now.Day(), 23, 59, 59, int(time.Second-time.Nanosecond), now.Location())
-
-	summaries, err := apiClient.Summaries(todayStart, todayEnd)
+	parsed, err := apiClient.Summary()
```
gandarez: it seems that `summary` or `today` makes more sense here.
```suggestion
	summary, err := apiClient.Summary()
```

pkg/api/summary.go
```diff
@@ -5,29 +5,24 @@ import (
...
 	"net/http"
-	"time"
 
 	"github.com/wakatime/wakatime-cli/pkg/summary"
 )
 
-const dateFormat = "2006-01-02"
-
-// Summaries fetches summaries for the defined date range.
+// Summary fetches code stats for Today.
```
gandarez: If the endpoint has changed to `/users/current/statusbar/today` also the file might be renamed to `today.go`. What do you think @dron22?

pkg/api/summary.go
```diff
@@ -5,29 +5,24 @@ import (
...
 //
 // ErrRequest is returned upon request failure with no received response from api.
 // ErrAuth is returned upon receiving a 401 Unauthorized api response.
 // Err is returned on any other api response related error.
-func (c *Client) Summaries(startDate, endDate time.Time) ([]summary.Summary, error) {
-	url := c.baseURL + "/users/current/summaries"
+func (c *Client) Summary() (*summary.Summary, error) {
```
gandarez: Maybe get rid of `summary`? 
```suggestion
func (c *Client) Today() (*summary.Summary, error) {
```

pkg/api/summary.go
```diff
@@ -56,59 +51,44 @@ func (c *Client) Summaries(startDate, endDate time.Time) ([]summary.Summary, err
 		))
 	}
 
-	summaries, err := ParseSummariesResponse(body)
+	summary, err := ParseSummaryResponse(body)
```
gandarez, line 54: We are parsing the result from `/users/current/statusbar/today`.
```suggestion
	today, err := ParseTodayResponse(body)
```

pkg/api/summary.go
```diff
@@ -56,59 +51,44 @@ func (c *Client) Summaries(startDate, endDate time.Time) ([]summary.Summary, err
...
 	}
 
-	return summaries, nil
+	return summary, nil
 }
 
-// ParseSummariesResponse parses the wakatime api response into summary.Summary.
-func ParseSummariesResponse(data []byte) ([]summary.Summary, error) {
+// ParseSummaryResponse parses the wakatime api response into summary.Summary.
+func ParseSummaryResponse(data []byte) (*summary.Summary, error) {
```
gandarez, line 63: See my comment above.

gandarez (changes requested):
I left some comments and suggestions. In general I'd replace all references to `summary` since we've replaced it by dedicated `today` endpoint and are not calling `summaries` endpoint by this PR.

cmd/legacy/today/today.go
alanhamlett: `summary` conflicts with a module name, so I'll use `today`.

pkg/api/summary.go
alanhamlett: Let's leave it `summary.Today` because the response is still a `summary.Summary` struct.

gandarez (approved):
let's get it merged and discuss better naming of changed packages. Also would be good to squash commits in this PR.

PR #533, opened 2021-09-05
Custom INI writer to prevent side effects

### LINKED ISSUE #532: go-ini writer with multiline values breaks legacy wakatime-cli for Vim
**Bug description**

If go-ini modifies any key in `~/.wakatime.cfg`, a side effect is every key gets parsed and re-written to the cfg file. If the cfg file contains a multiline value, that incorrectly gets written without indentation. Then when any legacy Python wakatime-cli reads the cfg file, it causes a crash:
```
Traceback (most recent call last):
  File "/wakatime/configs.py", line 58, in parseConfigFile
    configs.read_file(fh)
  File "/wakatime/packages/configparser/__init__.py", line 724, in read_file
    self._read(f, source)
  File "/wakatime/packages/configparser/__init__.py", line 1131, in _read
    raise e
ParsingError: Source contains parsing errors: '~/.wakatime.cfg'
	[line  3]: u'COMMIT_EDITMSG$\n'
	[line  4]: u'PULLREQ_EDITMSG$\n'
	[line  5]: u'MERGE_MSG$\n'
	[line  6]: u'TAG_EDITMSG$"""\n'
```

Example `~/.wakatime.cfg` file that causes the error:
```
[settings]
ignore             = """
COMMIT_EDITMSG$
PULLREQ_EDITMSG$
MERGE_MSG$
TAG_EDITMSG$"""
status_bar_enabled = true
debug=true
[internal]
cli_version = v1.22.1
cli_version_last_modified = Wed, 01 Sep 2021 21:48:38 GMT
backoff_retries           = 0
backoff_at                =
```

Example `~/.wakatime.cfg` file that does NOT cause the error:
```
[settings]
ignore             = """
 COMMIT_EDITMSG$
 PULLREQ_EDITMSG$
 MERGE_MSG$
 TAG_EDITMSG$"""
status_bar_enabled = true
debug=true
[internal]
cli_version = v1.22.1
cli_version_last_modified = Wed, 01 Sep 2021 21:48:38 GMT
backoff_retries           = 0
backoff_at                =
```

The only change is indenting the multiline `ignore` value with one space before each line.

alanhamlett:
Fixes #532.

### DIFF
--- a/pkg/backoff/backoff_internal_test.go
+++ b/pkg/backoff/backoff_internal_test.go
@@ -3,10 +3,12 @@ package backoff
 import (
 	"io/ioutil"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/wakatime/wakatime-cli/pkg/config"
+	"github.com/wakatime/wakatime-cli/pkg/ini"
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
@@ -50,16 +52,16 @@ func TestUpdateBackoffSettings(t *testing.T) {
 	err = updateBackoffSettings(v, 2, at)
 	require.NoError(t, err)
 
-	ini, err := config.NewIniWriter(v, func(vp *viper.Viper) (string, error) {
+	writer, err := config.NewIniWriter(v, func(vp *viper.Viper) (string, error) {
 		assert.Equal(t, v, vp)
 		return tmpFile.Name(), nil
 	})
 	require.NoError(t, err)
 
-	backoffAt := ini.File.Section("internal").Key("backoff_at").MustTimeFormat(config.DateFormat)
+	backoffAt := writer.File.Section("internal").Key("backoff_at").MustTimeFormat(config.DateFormat)
 
 	assert.WithinDuration(t, time.Now(), backoffAt, 15*time.Second)
-	assert.Equal(t, "2", ini.File.Section("internal").Key("backoff_retries").String())
+	assert.Equal(t, "2", writer.File.Section("internal").Key("backoff_retries").String())
 }
 
 func TestUpdateBackoffSettings_NotInBackoff(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/backoff/backoff_internal_test.go
```diff
@@ -75,12 +78,76 @@ func TestUpdateBackoffSettings_NotInBackoff(t *testing.T) {
...
+	actual, err := ioutil.ReadFile(tmpFile.Name())
+	require.NoError(t, err)
+
+	expected, err := ioutil.ReadFile("testdata/multiline-expected.cfg")
+	require.NoError(t, err)
+
+	assert.Equal(t, strings.ReplaceAll(string(expected), "\r", ""), string(actual))
+}
+
+func copyFile(t *testing.T, src string, dst string) {
```
gandarez: We already have a simplified `copyFile` function that copies from source to destination. It's implemented at `file_test.go`.

```Go
func copyFile(t *testing.T, source, destination string) {
	input, err := ioutil.ReadFile(source)
	require.NoError(t, err)

	err = ioutil.WriteFile(destination, input, 0600)
	require.NoError(t, err)
}
```

pkg/backoff/testdata/multiline.cfg
```diff
@@ -0,0 +1,8 @@
+[settings]
+ignore =
+ one
+ two
+debug = true
+[internal]
+backoff_retries = 1
+backoff_at = 2021-09-05T12:17:21-07:00
```
gandarez, line 8: Misssing break line at the end.

pkg/backoff/testdata/multiline-expected.cfg
```diff
@@ -0,0 +1,8 @@
+[settings]
+ignore =
+ one
+ two
+debug = true
+[internal]
+backoff_retries = 0
+backoff_at =
```
gandarez: Missing break line at the end.

pkg/ini/ini.go
```diff
@@ -0,0 +1,203 @@
...
+func SetKeys(iniFile string, values []Item) error {
+	if len(values) == 0 {
+		return nil
+	}
+
+	fh, err := os.Open(iniFile)
+	if err != nil {
+		return fmt.Errorf("failed to open ini file %q: %s", iniFile, err)
+	}
+
```
gandarez, line 104: Missing `defer fg.Close()` here.

pkg/ini/ini.go
```diff
@@ -0,0 +1,203 @@
...
+		if currentSection != value.Key.Section {
+			lines = append(lines, "["+value.Key.Section+"]")
+
+			currentSection = value.Key.Section
+		}
+
+		lines = append(lines, value.Key.Name+" = "+value.Value)
+	}
+
+	fh.Close()
```
gandarez, line 178: Might be replaced by defer after its assigning.

gandarez (changes requested):
Left some comments. Would be extremely important to have unit tests for ini package and also we use underscore to separate words in file names (`multiline-expected.cfg`).

pkg/ini/ini.go
alanhamlett: I can't use `defer fh.Close()` because we manually call `Close()` before re-opening the file to write it's new contents. `defer` isn't necessary here because we only have a single `return` at the end of the function.

PR #537, opened 2021-09-06
Add entity modifier middleware

gandarez:
This PR replaces the conditional validation for XCode Playground to a more flexible and testable function. It adds a new middleware called `Entity Modifier` that is intended to be used when changes are required to Entity's path.

In addition this PR bumps Go to 1.17.


alanhamlett (approved):

alanhamlett:
👍

PR #538, opened 2021-09-07
Handle status code 400

gandarez:
This PR handles http status code 400 (Bad Request) and return default exit (1).

### DIFF
--- a/cmd/legacy/heartbeat/heartbeat.go
+++ b/cmd/legacy/heartbeat/heartbeat.go
@@ -42,11 +42,19 @@ func Run(v *viper.Viper) (int, error) {
 			)
 		}
 
+		var errbadRequest api.ErrBadRequest
+		if errors.As(err, &errbadRequest) {
+			return exitcode.ErrDefault, fmt.Errorf(
+				"failed to send heartbeat(s) due to api error: %s",
+				errbadRequest,
+			)
+		}
+
 		var errapi api.Err
 		if errors.As(err, &errapi) {
 			return exitcode.ErrAPI, fmt.Errorf(
 				"failed to send heartbeat(s) due to api error: %s",
-				err,
+				errbadRequest,
 			)
 		}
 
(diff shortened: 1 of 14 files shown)


alanhamlett (commented):
We need integration tests, because `errors.As(err, &errapi)` etc aren't matching error types correctly. Maybe because the error type is handled and re-returned as a different type before it gets to the `errors.As` line.

Either way, we need tests that check the exit code.

alanhamlett:
I think this is the line that breaks matching error type, by re-returning the error as the wrong type:

https://github.com/wakatime/wakatime-cli/blob/0760642ff424b8a9402b5dfa9e498825ffc0fd69/cmd/legacy/heartbeat/heartbeat.go#L187

gandarez:
@alanhamlett added an integration test to validate error when "bad request".

alanhamlett:
> added an integration test to validate error when "bad request".

In the tests, we should make sure the exit code matches `exitcode.ErrAPI` or for `400` that it matches `exitcode.ErrDefault`.

gandarez:
> > added an integration test to validate error when "bad request".
> 
> In the tests, we should make sure the exit code matches `exitcode.ErrAPI` or for `400` that it matches `exitcode.ErrDefault`.

Integration test now checks for exit code.

cmd/legacy/heartbeat/heartbeat.go
```diff
@@ -42,11 +42,19 @@ func Run(v *viper.Viper) (int, error) {
 			)
 		}
 
+		var errbadRequest api.ErrBadRequest
+		if errors.As(err, &errbadRequest) {
+			return exitcode.ErrDefault, fmt.Errorf(
+				"failed to send heartbeat(s) due to api error: %s",
```
alanhamlett, line 48: will this result in a log line with repeated text? `failed to send heartbeat(s) due to api error: failed to send heartbeat(s) due to api error: ...`

alanhamlett (approved):

PR #539, opened 2021-09-07
All api errors except 400 and 401 should exit with 102 exit code

alanhamlett:
This prevents the IDE's status bar from showing `WakaTime Error` when working offline or having temporary connection issues, as long as the heartbeat was saved to the local offline db.

### DIFF
--- a/cmd/legacy/offlinesync/offlinesync.go
+++ b/cmd/legacy/offlinesync/offlinesync.go
@@ -18,7 +18,7 @@ import (
 func Run(v *viper.Viper) (int, error) {
 	queueFilepath, err := offline.QueueFilepath()
 	if err != nil {
-		return exitcode.ErrDefault, fmt.Errorf(
+		return exitcode.ErrGeneric, fmt.Errorf(
 			"failed to load offline queue filepath: %s",
 			err,
 		)
@@ -29,15 +29,15 @@ func Run(v *viper.Viper) (int, error) {
 		var errauth api.ErrAuth
 		if errors.As(err, &errauth) {
 			return exitcode.ErrAuth, fmt.Errorf(
-				"failed to sync offline activity: %s. Find your api key from wakatime.com/settings/api-key",
+				"invalid api key... find yours at wakatime.com/settings/api-key. %s",
 				errauth,
 			)
 		}
 
-		var errbadRequest api.ErrBadRequest
+		var errbadRequest api.Err
 		if errors.As(err, &errbadRequest) {
-			return exitcode.ErrDefault, fmt.Errorf(
-				"failed to sync offline activity due to api error: %s",
+			return exitcode.ErrGeneric, fmt.Errorf(
+				"unable to sync offline activity due to api error: %s",
 				err,
 			)
 		}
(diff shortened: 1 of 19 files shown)


pkg/api/error.go
```diff
@@ -16,18 +16,10 @@ func (e ErrAuth) Error() string {
...
-// ErrBadRequest represents a bad request error.
+// ErrBadRequest represents a 400 response from the API.
 type ErrBadRequest string
 
 // Error method to implement error interface.
 func (e ErrBadRequest) Error() string {
 	return string(e)
 }
-
-// ErrRequest represents a request failure, where no response was received from the api.
```
alanhamlett, line 27: Removed because redundant. All request errors should be handled the same except `ErrBadRequest (400)` and `ErrAuth (401)` responses.

cmd/legacy/offlinesync/offlinesync.go
```diff
@@ -29,29 +29,29 @@ func Run(v *viper.Viper) (int, error) {
 		var errauth api.ErrAuth
 		if errors.As(err, &errauth) {
 			return exitcode.ErrAuth, fmt.Errorf(
-				"failed to sync offline activity: %s. Find your api key from wakatime.com/settings/api-key",
+				"invalid api key... find yours at wakatime.com/settings/api-key. %s",
```
gandarez, line 32: We'll miss tracking where the error happened. The upstream won't be able to reference "offline" function in the logs.

cmd/legacy/today/today.go
```diff
@@ -21,28 +21,28 @@ func Run(v *viper.Viper) (int, error) {
 		var errauth api.ErrAuth
 		if errors.As(err, &errauth) {
 			return exitcode.ErrAuth, fmt.Errorf(
-				"failed to fetch today: %s. Find your api key from wakatime.com/settings/api-key",
+				"invalid api key... find yours at wakatime.com/settings/api-key. %s",
```
gandarez, line 24: We'll miss tracking where the error happened.

cmd/legacy/todaygoal/todaygoal.go
```diff
@@ -30,28 +30,28 @@ func Run(v *viper.Viper) (int, error) {
 		var errauth api.ErrAuth
 		if errors.As(err, &errauth) {
 			return exitcode.ErrAuth, fmt.Errorf(
-				"failed to fetch today goal: %s. Find your api key from wakatime.com/settings/api-key",
+				"invalid api key... find yours at wakatime.com/settings/api-key. %s",
```
gandarez, line 33: We'll miss tracking where the error happened.

cmd/legacy/offlinesync/offlinesync.go
alanhamlett: Doesn't it concat that together with `command failed: offline`?

alanhamlett: Oh you're right, it results in log:

```
"message":"failed to run command: invalid api key... find yours at wakatime.com/settings/api-key. authentication failed at...
```

I'll fix this in a new PR.

PR #547, opened 2021-09-17
Fallback to getpwuid when HOME env var not set

alanhamlett:
Let's see if this can detect the current user's home folder when `$HOME` env var is unset. Should fix https://github.com/wakatime/vscode-wakatime/pull/238.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,9 +33,11 @@ else
 endif
 
 # targets
-build-all: build-darwin build-freebsd build-linux build-netbsd build-openbsd build-windows
+build-all:
+	build-darwin build-freebsd build-linux build-netbsd build-openbsd build-windows
 
-build-darwin: build-darwin-amd64 build-darwin-arm64
+build-darwin:
+	build-darwin-amd64 build-darwin-arm64
 
 build-darwin-amd64:
 	GOOS=darwin GOARCH=amd64 make build-binary
@@ -54,7 +56,8 @@ build-freebsd-amd64:
 build-freebsd-arm:
 	GOOS=freebsd GOARCH=arm make build-binary
 
-build-linux: build-linux-386 build-linux-amd64 build-linux-arm build-linux-arm64
+build-linux:
+	build-linux-386 build-linux-amd64 build-linux-arm build-linux-arm64
 
 build-linux-386:
 	GOOS=linux GOARCH=386 make build-binary
@@ -68,7 +71,8 @@ build-linux-arm:
 build-linux-arm64:
 	GOOS=linux GOARCH=arm64 make build-binary
 
-build-netbsd: build-netbsd-386 build-netbsd-amd64 build-netbsd-arm
+build-netbsd:
+	build-netbsd-386 build-netbsd-amd64 build-netbsd-arm
 
 build-netbsd-386:
 	GOOS=netbsd GOARCH=386 make build-binary
@@ -79,7 +83,8 @@ build-netbsd-amd64:
 build-netbsd-arm:
 	GOOS=netbsd GOARCH=arm make build-binary
 
-build-openbsd: build-openbsd-386 build-openbsd-amd64 build-openbsd-arm build-openbsd-arm64
+build-openbsd:
+	build-openbsd-386 build-openbsd-amd64 build-openbsd-arm build-openbsd-arm64
 
 build-openbsd-386:
 	GOOS=openbsd GOARCH=386 make build-binary
@@ -102,7 +107,7 @@ build-windows-amd64:
 	GOOS=windows GOARCH=amd64 make build-binary-windows
 
 build-binary:
-	CGO_ENABLED="0" GOOS=$(GOOS) GOARCH=$(GOARCH) $(GOBUILD) -v \
+	CGO_ENABLED="1" GOOS=$(GOOS) GOARCH=$(GOARCH) $(GOBUILD) -v \
 		-ldflags "${LD_FLAGS} -X ${REPO}/pkg/version.OS=$(GOOS) -X ${REPO}/pkg/version.Arch=$(GOARCH)" \
 		-o ${BUILD_DIR}/$(BINARY_NAME)-$(GOOS)-$(GOARCH)
 
--- a/cmd/legacy/logfile/logfile.go
+++ b/cmd/legacy/logfile/logfile.go
@@ -2,9 +2,9 @@ package logfile
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 
+	"github.com/wakatime/wakatime-cli/pkg/config"
 	"github.com/wakatime/wakatime-cli/pkg/vipertools"
 
 	"github.com/mitchellh/go-homedir"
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -48,8 +48,7 @@ build-freebsd: build-freebsd-386 build-freebsd-amd64 build-freebsd-arm
 build-freebsd-386:
 	GOOS=freebsd GOARCH=386 make build-binary
 
-build-freebsd-amd64:
-	GOOS=freebsd GOARCH=amd64 make build-binary
+build-freebsd-amd64: GOOS=freebsd GOARCH=amd64 make build-binary
```
gandarez: Is it really necessary? All the others have a line break.

cmd/legacy/logfile/logfile.go
```diff
@@ -67,3 +53,21 @@ func LoadParams(v *viper.Viper) (Params, error) {
 		Verbose:  v.GetBool("verbose") || debug,
 	}, nil
 }
+
+func findHomeDir() string {
+	home, exists := os.LookupEnv("WAKATIME_HOME")
+	if exists && home != "" {
+		home, err := homedir.Expand(home)
+		if err == nil {
```
gandarez: In Go we usually check for `err != nil`. This func could be refactored to
```Go
func findHomeDir() string {
	home, exists := os.LookupEnv("WAKATIME_HOME")
	if exists && home != "" {
		home, err := homedir.Expand(home)
		if err != nil {
			goto fallback
		}

		return home
	}

fallback:
	home, err := os.UserHomeDir()
	if err != nil {
		usr := pwd.Getpwuid(uint32(os.Getuid()))
		return usr.Dir
	}

	return home
}
```

pkg/api/transport_windows.go
```diff
@@ -1,4 +1,4 @@
-// +build windows
+//go:build windows
```
gandarez, line 1: 👍

pkg/pwd/pwd_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+		UID:    uint32(cpw.pw_uid),
+		GID:    uint32(cpw.pw_uid),
+		Gecos:  C.GoString(cpw.pw_gecos),
+		Dir:    C.GoString(cpw.pw_dir),
+		Shell:  C.GoString(cpw.pw_shell),
+	}
+}
+
+// Getpwnam searches the user database for an entry with a matching name.
+func Getpwnam(name string) *Passwd {
```
gandarez: In Go we avoid using `Get` as prefixes for getter functions. In this case We'd rename to `Pwnam`. It's better explained [here](https://golang.org/doc/effective_go#Getters). Should do the same for the other functions.

pkg/pwd/pwd_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+	cname := C.CString(name)
+
+	defer C.free(unsafe.Pointer(cname))
+
+	cpw := C.getpwnam(cname)
+	if cpw == nil {
```
gandarez:
```suggestion
if cpw != nil {
	return cpasswd2go(cpw)
}

return nil
```

pkg/pwd/pwd_cgo.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil
+	}
+
+	return cpasswd2go(cpw)
+}
+
+// Getpwuid searches the user database for an entry with a matching uid.
+func Getpwuid(uid uint32) *Passwd {
+	cpw := C.getpwuid_aux(C.uint(uid))
+	if cpw == nil {
```
gandarez:
```suggestion
if cpw != nil {
	return cpasswd2go(cpw)
}

return nil
```

pkg/pwd/pwd_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+package pwd_test
+
+import (
+	"testing"
+
+	"github.com/wakatime/wakatime-cli/pkg/pwd"
+)
+
+func TestGetpwnam(t *testing.T) {
```
gandarez: For multiple scenarios would be better to write a `map[string]struct{...}` likewise we've done for others to test them.

```Go
func TestGetpwnam(t *testing.T) {
	tests := map[string]struct {
		Name     string
		Expected *string
	}{
		"root": {
			Name:     "root",
			Expected: String("root"),
		},
		"nobody": {
			Name:     "nobody",
			Expected: String("nobody"),
		},
		"not existing": {
			Name:     "not existing",
			Expected: nil,
		},
	}

	for name, test := range tests {
		t.Run(name, func(t *testing.T) {
			res := pwd.Getpwnam(test.Name)

			if test.Expected != nil {
				assert.Equal(t, test.Expected, &res.Name)
			} else {
				assert.Nil(t, res)
			}
		})
	}
}
```

pkg/config/config.go
```diff
@@ -104,20 +105,30 @@ func FilePath(v *viper.Viper) (string, error) {
...
+	home := WakaHomeDir()
+	if home == "" {
+		return "", errors.New("failed getting user's home directory")
+	}
+
+	return filepath.Join(home, defaultFile), nil
+}
+
+// WakaHomeDir returns the current user's home directory.
+func WakaHomeDir() string {
```
gandarez, line 117: In Go we check for `err != nil` in most scenarios. In this case we'd refactor to something like this, what do you think?

```Go
unc WakaHomeDir() string {
	home, exists := os.LookupEnv("WAKATIME_HOME")
	if exists && home != "" {
		home, err := homedir.Expand(home)
		if err != nil {
			goto fallback
		}

		return home
	}

fallback:
	home, err := os.UserHomeDir()
	if err != nil {
		return pwd.Getpwuid(uint32(os.Getuid())).Dir
	}

	return home
}
```

pkg/config/config.go
```diff
@@ -104,20 +105,30 @@ func FilePath(v *viper.Viper) (string, error) {
...
 
 	home, err := os.UserHomeDir()
-	if err != nil {
-		return "", fmt.Errorf("failed getting user's home directory: %s", err)
+	if err == nil {
+		return home
 	}
 
-	return filepath.Join(home, defaultFile), nil
+	usr := pwd.Getpwuid(uint32(os.Getuid()))
```
gandarez: No need to create a new var and return.

gandarez (changes requested):
Some lines of code might be changed to be more "Go style".

Makefile
alanhamlett: VSCode did that automatically

alanhamlett: Fixed.

pkg/pwd/pwd_cgo.go
alanhamlett: This is just named the same as the C function.

gandarez: I got it but in Go we should avoid `Get` as prefixes. The package gives the context and the function is expected to always return something.

pkg/pwd/pwd_cgo.go
```diff
@@ -0,0 +1,56 @@
...
+	Name   string // user name
+	Passwd string // user password
+	UID    uint32 // user ID
+	GID    uint32 // group ID
+	Gecos  string // real name
+	Dir    string // home directory
+	Shell  string // shell program
+}
+
+func cpasswd2go(cpw *C.struct_passwd) *Passwd {
```
gandarez: Why do we need a private func called only once and that only returns a pointer to a struct?

gandarez (changes requested):

gandarez (approved):
lgtm

pkg/pwd/pwd_cgo.go
alanhamlett: Removed 👍

PR #563, opened 2021-09-30
Use branchlint-action at master

alanhamlett:


gandarez:
That's not `semver-action` right? We're changing for `branchlint-action`. I personally don't like this approach of using master ones from 3rd parties since them can easily introduce a new bug and it will also reflect on us.

alanhamlett:
Yes, sorry I misread the repo name. We should keep external dependencies versioned.

PR #584, opened 2021-11-20
Internal config parse

gandarez:
This PR moves the responsability of loading `internal-config` file to the main entry likewise for reading the regular config file and also gives back the initial approach of reading any param using Viper. It was also breaking tests on my side because there was backoff set at my waka config file. In addition it also includes a fix for filtering heartbeat where the error message was truncated.

### DIFF
--- a/cmd/legacy/heartbeat/heartbeat.go
+++ b/cmd/legacy/heartbeat/heartbeat.go
@@ -92,6 +92,10 @@ func RunWithoutSending(v *viper.Viper) (int, error) {
 	handleOpts := initHandleOptions(params)
 
 	if !params.OfflineDisabled {
+		if params.OfflineQueueFile != "" {
+			queueFilepath = params.OfflineQueueFile
+		}
+
 		offlineHandleOpt, err := offline.WithQueue(queueFilepath)
 		if err != nil {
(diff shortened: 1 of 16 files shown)


dron22:
Hey @gandarez! I will review this later today at about 8pm CET  :+1:

dron22 (approved):
Nice one :+1: Great that we read everything via VIper now :rocket: 

Should also be reviewed by @alanhamlett, as I was not really involved, when we added this internal config file support.

alanhamlett:
LGTM... I manually parsed the internal ini file because I couldn't find a way to parse two config files with the same Viper instance.

alanhamlett:
Maybe out of scope in this PR, but can we add a test for corrupted internal config file and make sure the heartbeat is saved to offline db?

Something like:
```ini
[internal
backoff_at = abc
```

(missing closing square bracket)

gandarez:
I don't think it's out of scope. I'll add a test for malformed internal config file.

gandarez:
I've added two new integration tests where one for regular malformed config and another one for malformed internal config. I also fixed reading multiple configs using Viper by changing from `ReadInConfig` to `MergeInConfig` and added a test for that.

alanhamlett (approved):

PR #586, opened 2021-11-23
Fix parsing extra heartbeats

### LINKED ISSUE #583: Parsing extra param int as str always fails
When parsing extra params json fails because an int was jsonified as str, a second attempt is made with a different struct accepting str. For example, when `cursorpos` is a string instead of int:

`{"timestamp": 123.456, "cursorpos": "42"}`

But then parsing timestamp fails because it's not rendered as string:

```
"failed to read extra heartbeats: failed to json decode: failed to json decode from data \"[{\\\"cursorpos\\\": \\\"42\\\", \\\"timestamp\\\": 1637185415.093967}]\\n\": json: cannot unmarshal string into Go struct field .cursorpos of type int: failed to json decode again, accepting string values: failed to json decode from data \"[{\\\"cursorpos\\\": \\\"42\\\", \\\"timestamp\\\": 1637185415.093967}]\\n\": json: cannot unmarshal number into Go struct field .timestamp of type string"
```

This all-or-nothing approach is too brittle. Instead, we must be able to parse any int or float param as either string or number without errors.

gandarez:
This PR fixes extra heartbeats parsing to accept any kind of data. Strong types were replaced by `interface{}` in order to be more flexible in terms of received json.

fixes https://github.com/wakatime/wakatime-cli/issues/583

### DIFF
--- a/cmd/legacy/heartbeat/params.go
+++ b/cmd/legacy/heartbeat/params.go
@@ -284,19 +284,9 @@ func readExtraHeartbeats() ([]heartbeat.Heartbeat, error) {
 		return nil, fmt.Errorf("failed to read data from stdin: %s", err)
 	}
 
-	heartbeats, errFirst := parseExtraHeartbeat(input)
-	if errFirst == nil {
-		return heartbeats, nil
-	}
-
-	// try again accepting string properties for int values
-	heartbeats, errSecond := parseExtraHeartbeatWithStringValues(input)
-	if errSecond != nil {
-		return nil, fmt.Errorf(
-			"failed to json decode: %s: failed to json decode again, accepting string values: %s",
-			errFirst,
-			errSecond,
-		)
+	heartbeats, err := parseExtraHeartbeat(input)
+	if err != nil {
+		return nil, fmt.Errorf("failed to json decode: %s", err)
 	}
 
 	return heartbeats, nil
@@ -305,19 +295,19 @@ func readExtraHeartbeats() ([]heartbeat.Heartbeat, error) {
 func parseExtraHeartbeat(data string) ([]heartbeat.Heartbeat, error) {
 	var incoming []struct {
 		Category          heartbeat.Category `json:"category"`
-		CursorPosition    *int               `json:"cursorpos"`
+		CursorPosition    interface{}        `json:"cursorpos"`
 		Entity            string             `json:"entity"`
 		EntityType        string             `json:"entity_type"`
 		Type              string             `json:"type"`
-		IsWrite           *bool              `json:"is_write"`
+		IsWrite           interface{}        `json:"is_write"`
 		Language          *string            `json:"language"`
 		LanguageAlternate string             `json:"alternate_language"`
-		LineNumber        *int               `json:"lineno"`
-		Lines             *int               `json:"lines"`
+		LineNumber        interface{}        `json:"lineno"`
+		Lines             interface{}        `json:"lines"`
 		Project           string             `json:"project"`
 		ProjectAlternate  string             `json:"alternate_project"`
-		Time              float64            `json:"time"`
-		Timestamp         float64            `json:"timestamp"`
+		Time              interface{}        `json:"time"`
+		Timestamp         interface{}        `json:"timestamp"`
 	}
 
 	err := json.Unmarshal([]byte(data), &incoming)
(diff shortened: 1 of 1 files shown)


dron22:
Hey @gandarez! I will review this later today at about 8pm CET :+1:

cmd/legacy/heartbeat/params.go
```diff
@@ -305,19 +295,19 @@ func readExtraHeartbeats() ([]heartbeat.Heartbeat, error) {
 func parseExtraHeartbeat(data string) ([]heartbeat.Heartbeat, error) {
 	var incoming []struct {
 		Category          heartbeat.Category `json:"category"`
-		CursorPosition    *int               `json:"cursorpos"`
+		CursorPosition    interface{}        `json:"cursorpos"`
 		Entity            string             `json:"entity"`
 		EntityType        string             `json:"entity_type"`
 		Type              string             `json:"type"`
-		IsWrite           *bool              `json:"is_write"`
```
alanhamlett, line 312: What values will be accepted as true/false?

gandarez: Any bool or string. `true`, `false`, `"true"`and `"false"`.

cmd/legacy/heartbeat/params.go
```diff
@@ -344,128 +334,97 @@ func parseExtraHeartbeat(data string) ([]heartbeat.Heartbeat, error) {
...
-
-	var heartbeats []heartbeat.Heartbeat
-
-	for _, h := range incoming {
 		var cursorPosition *int
 
-		if h.CursorPosition != nil {
-			parsed, err := strconv.Atoi(*h.CursorPosition)
+		switch cursorPositionVal := h.CursorPosition.(type) {
+		case float64:
```
dron22, line 340: I guess we should add a case for `int` type as well. This way, we support `float64`, `int` and `string`. This applies for `lineno` and `lines` values as well.

cmd/legacy/heartbeat/params.go
```diff
@@ -344,128 +334,97 @@ func parseExtraHeartbeat(data string) ([]heartbeat.Heartbeat, error) {
...
 		}
 
-		var timestamp float64
+		var time float64
 
-		switch {
-		case h.Time != nil:
-			parsed, err := strconv.ParseFloat(*h.Time, 64)
+		switch timeVal := h.Time.(type) {
+		case float64:
```
dron22, line 396: For `time` property, support for case `int` type would make it more robust as well.

cmd/legacy/heartbeat/params.go
```diff
@@ -344,128 +334,97 @@ func parseExtraHeartbeat(data string) ([]heartbeat.Heartbeat, error) {
...
-			timestamp = parsed
-		case h.Timestamp != nil:
-			parsed, err := strconv.ParseFloat(*h.Timestamp, 64)
+			time = val
+		}
+
+		var timestamp float64
+
+		switch timestampVal := h.Timestamp.(type) {
+		case float64:
```
dron22, line 410: For `timestamp` property, support for case `int` type would make it more robust as well.

dron22 (dismissed):
In general LGTM, nice usage of type switches :+1: To make it completely robust, I think we need additional cases for `int` types.

In general adding test cases to cover this new behavior would be good.

gandarez:
I think we've covered all use cases for extra heartbeats which are string or the type itself (`TestLoadParams_ExtraHeartbeats` and `TestLoadParams_ExtraHeartbeats_WithStringValues`). When Go unmarshals for `interface{}` then it's converted to the following types under the hood. It's documented [here](https://pkg.go.dev/encoding/json#Unmarshal). So we actually don't need to add an extra case to verify `int` type when parsing for numbers because it'll always be `float64`.

```
bool, for JSON booleans
float64, for JSON numbers
string, for JSON strings
[]interface{}, for JSON arrays
map[string]interface{}, for JSON objects
nil for JSON null
```

alanhamlett:
> `float64, for JSON numbers`

So it always uses `float64` when parsing JSON int and floats?

alanhamlett:
Let's add tests for all possible inputs and then LGTM!

gandarez:
All possible inputs are already covered by `TestLoadParams_ExtraHeartbeats` and `TestLoadParams_ExtraHeartbeats_WithStringValues`. Do you guys see other ways to tests?

alanhamlett:
> All possible inputs are already covered by `TestLoadParams_ExtraHeartbeats` and `TestLoadParams_ExtraHeartbeats_WithStringValues`. Do you guys see other ways to tests?

We could change `"lineno": "42"` to `"lineno": 42` but only in one of the two heartbeats here:
https://github.com/wakatime/wakatime-cli/blob/970e80e2139dcff9e772bf79eab4fb9046f6f790/cmd/legacy/heartbeat/testdata/extra_heartbeats_with_string_values.json#L1

That would test for the actual bug in #583 which is triggered from mixing string and integer values in a heartbeat.

gandarez:
@alanhamlett because we changed from strong types to interfaces and we do have tests for both I don't see any advantage of changing one of them from string to the real type. Those scenarios are already covered because in the past we had two independent structs, where one for the strong types and only accepted real values and another one to accept only strings. The new scenario fixes this approach and accepts any kind of value. We need to make sure we're testing both cases for each (interfaced) field.

alanhamlett (approved):

alanhamlett:
Yea, that makes sense. In that case, looks great!

PR #587, opened 2021-11-26
Release v1.28.6

gandarez:
Changelog:
- Fix parsing extra heartbeats #583 
- Fix internal config file parse
- Set very large 10 minute timeout when opening boltdb file, to prevet waiting infinitely


alanhamlett (approved):
🚀

PR #588, opened 2021-11-28
Revert to use gopkg.in/ini.v1 when saving config file

gandarez:
This PR adds back `gopkg.in/ini.v1` and removes `pkg/ini` custom ini parser. The lib now won't change the indentation of multiline python strings.

For further reference please see this [PR](https://github.com/go-ini/ini/pull/299).


alanhamlett (approved):

alanhamlett:
Looks like we need two testdata files, one for Windows and one for Linux/Mac, because the library is writing the Windows version with `\r\n` line endings.

PR #590, opened 2021-11-29
Lock config file to prevent corruption

### LINKED ISSUE #576: Lock config files when reading/writing to prevent corruption
Some recent bugs (wakatime/jetbrains-wakatime#192 and wakatime/vscode-wakatime#232) suggest that the `~/.wakatime.cfg` and `~/.wakatime-internal.cfg` files may become corrupted if two wakatime-cli processes write to the file at the same time. This is only a guess, so maybe something else is causing the files to become corrupted. However, it can't hurt to lock anytime we read/write those files to prevent clobbering them.

gandarez:
This PR adds locking to config file when writing to prevent corruption by concurrency.

Fixes #576


cmd/legacy/configread/configread.go
```diff
@@ -27,7 +27,7 @@ func Run(v *viper.Viper) (int, error) {
 		)
 	}
 
-	fmt.Println(output)
+	fmt.Print(output)
```
alanhamlett: won't this cause the output to be joined with the terminal prompt if output is the last line of the config file without a newline? How about we strip newlines from end of output, then print with a newline?

gandarez: We actually don't need to change it. Even for last line where there's no newline it will print accorndingly.We actually don't need to change it. Even for last line where there's no newline it will print accorndingly.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/alecthomas/chroma v0.8.2
 	github.com/danwakefield/fnmatch v0.0.0-20160403171240-cbb64ac3d964
 	github.com/dlclark/regexp2 v1.4.0
+	github.com/juju/fslock v0.0.0-20160525022230-4d5c94c67b4b
```
alanhamlett: There's a discussion about [deleting the fslock repo](https://github.com/juju/fslock/issues/4).

gandarez: I've seen that but honestly I don't think they will. Let's change to another lib and see if it works.

pkg/config/config.go
```diff
@@ -65,6 +68,20 @@ func (w *IniWriter) Write(section string, keyValue map[string]string) error {
 		w.File.Section(section).Key(key).SetValue(value)
 	}
 
+	lock := fslock.New(w.ConfigFilepath)
```
alanhamlett: Should we add a suffix `.lock` to `ConfigFilepath`? I think fslock expects the file to be a tmp file, not a file we want to read/write in the same process after locking.

go.mod
gandarez: What about forking it to prevent further deleting?

gandarez:
@alanhamlett I've made it much simpler by using Mutex.

alanhamlett:
`juju/mutex` is missing from `go.mod`.

`sync.mutex` is only for locking within the same process, we should use https://github.com/juju/mutex.

alanhamlett (changes requested):
We should use [juju/mutex](https://github.com/juju/mutex) instead of [sync.Mutex](https://pkg.go.dev/sync#Mutex) because we want multi-process locking mainly on the internal config file.

main_test.go
```diff
@@ -512,13 +543,11 @@ func runCmdExpectErr(cmd *exec.Cmd) (string, int) {
...
 	cmd.Stderr = &stderr
 
-	version.Version = testVersion
-
 	err := cmd.Run()
 	if err == nil {
 		fmt.Println(stdout.String())
 		fmt.Println(stderr.String())
-		fmt.Printf("ran command successfully, but was expecting error: %s\n", cmd)
+		fmt.Printf("ran command successfully, but was expecting error: %s: %s\n", cmd, err)
```
dron22: err will always be `nil` here.

dron22 (commented):
I only see test changes in `main_test.go`. Am I missing something?

main_test.go
gandarez: good catch

gandarez:
@dron22 it was missing the last implementation. @alanhamlett I'm not too sure about the configuration values required for mutex to work but IMHO it looks good.

pkg/config/config.go
```diff
@@ -157,3 +176,16 @@ func WakaHomeDir() (string, error) {
 
 	return "", allerrs
 }
+
+// mutexClock is used to implement mutex.Clock interface.
+type mutexClock struct {
+	delay time.Duration
+}
+
+func (mc *mutexClock) After(time.Duration) <-chan time.Time {
```
alanhamlett, line 185: Is this used?

alanhamlett (approved):
Looks good! `juju/mutex` will prevent two wakatime-cli processes from writing to the config file at the same time 👍

pkg/config/config.go
gandarez: Yep it's needed because `mutex.Spec` expects a `Clock`interface implementaion.

pkg/config/config.go
```diff
@@ -65,6 +68,22 @@ func (w *IniWriter) Write(section string, keyValue map[string]string) error {
...
 	}
 
+	releaser, err := mutex.Acquire(mutex.Spec{
+		Name:    "wakatime-cli-config-mutex",
+		Delay:   time.Millisecond,
+		Timeout: defaultTimeout,
+		Clock:   &mutexClock{delay: time.Millisecond},
+	})
+	if err != nil {
+		log.Debugf("failed to acquire mutex: %s", err)
```
alanhamlett, line 78: This is perfect, in case something goes wrong we should still try and continue writing the configs and saving heartbeats to offline db 👍

PR #598, opened 2021-12-10
Improve project obfuscation durability

### LINKED ISSUE #575: cli creates .wakatime-project in root
Cli creates `.wakatime-project` in the root directory (`~/` in my case) which results in all of my projects being pushed tagged with the generated name provided in `~/.wakatime-project`.

Create `.wakatime-project` only in specified paths in `hide_project_names` flag
It works but also generates a project name in root.
**Environment**:
Ubuntu 20.04.3 LTS
wakatime version `13.1.0`, python I guess

**Logs**:
Secret info but for the date when `~/.wakatime-project` file was created there are only errors about `failed to run command: sending heartbeat(s) later due to api error: won't send heartbeat due to backoff`

My config
```cfg
[settings]
api_key = {secret}
status_bar_coding_activity = true
status_bar_enabled = true
hide_file_names = 
    ^/home/nickname/git/private
    ^/home/nickname/job
hide_project_names = 
    ^/home/nickname/git/private
    ^/home/nickname/job
```

alanhamlett:
Maybe this will fix #575.


pkg/project/project.go
```diff
@@ -219,29 +235,41 @@ func generateProjectName() string {
...
-			return filepath.Join(startDir, fileDir, filename), true
+		if isRootPath(directory) {
+			return "", false
 		}
 
-		startDir = filepath.Clean(filepath.Join(startDir, ".."))
-		if startDir == "." || startDir == "/" || driveLetterRegex.MatchString(startDir) {
-			return "", false
+		// remove any trailing file from directory
+		directory = filepath.Dir(directory + string(filepath.Separator))
```
alanhamlett: Previously the first iteration would skip the current directory, for ex:

```go
 FindFileOrDirectory("/path/to/file", ".wakatime-project")
```

Would only start looking for `.wakatime-project` at `/path/.wakatime-project` missing any file at `/path/to/.wakatime-project`. This might have caused the bug seen in #575.

cmd/legacy/heartbeat/params.go
```diff
@@ -623,20 +623,20 @@ func parseBoolOrRegexList(s string) ([]regex.Regex, error) {
 	var patterns []regex.Regex
 
 	switch {
-	case s == "":
+	case strings.Trim(s, "\n\r\t ") == "":
```
gandarez: why not to trim before switching?

pkg/project/project.go
```diff
@@ -16,6 +18,10 @@ import (
 // nolint: gochecknoglobals
 var driveLetterRegex = regexp.MustCompile(`^[a-zA-Z]:\\$`)
 
+// WakaTimeProjectFile is the special file which if present should contain the project name and optional branch name
+// that will be used instead of the auto-detected project and branch names.
+const WakaTimeProjectFile = ".wakatime-project"
```
gandarez: let's aggregate with `const ( ... )`.

pkg/project/project.go
gandarez: Not sure it would happen because of current implementation. Did you get this behavior on tests?

pkg/project/project.go
```diff
@@ -219,29 +235,41 @@ func generateProjectName() string {
...
+			return filepath.Join(directory, filename), true
 		}
 
+		directory = filepath.Clean(filepath.Join(directory, ".."))
+
 		i++
 	}
 
-	log.Warnf("didn't find %s after %d iterations", filename, maxRecursiveIteration)
+	log.Warnf("Max %d iterations reached without finding %s", maxRecursiveIteration, filename)
```
gandarez: log should be lower case

pkg/project/project.go
```diff
@@ -219,29 +235,41 @@ func generateProjectName() string {
...
 
 	return "", false
 }
 
+func isRootPath(directory string) bool {
```
gandarez, line 263: Don't think `isRootPath` is needed because it's only used once.

pkg/project/subversion.go
```diff
@@ -39,12 +39,12 @@ func (s Subversion) Detect() (Result, bool, error) {
...
 
 	// Find for .svn/wc.db file
-	svnConfigFile, ok := FindFileOrDirectory(fp, ".svn", "wc.db")
+	svnConfigFile, ok := FindFileOrDirectory(fp, filepath.Join(".svn", "wc.db"))
 	if !ok {
 		return Result{}, false, nil
 	}
 
-	info, ok, err := svnInfo(filepath.Join(svnConfigFile, "../.."), binary)
+	info, ok, err := svnInfo(filepath.Join(svnConfigFile, "..", ".."), binary)
```
gandarez, line 47: 👍

gandarez (changes requested):

pkg/project/project.go
alanhamlett: It's for readability... it makes the code using `isRootPath` more readable.

gandarez: what about decorating with a comment?

cmd/legacy/heartbeat/params.go
```diff
@@ -622,21 +622,24 @@ func loadProjectParams(v *viper.Viper) (ProjectParams, error) {
...
+	case s == "":
 		break
-	case strings.ToLower(strings.Trim(s, "\n\r\t ")) == "false":
+	case strings.ToLower(s) == "false":
 		break
-	case strings.ToLower(strings.Trim(s, "\n\r\t ")) == "true":
+	case strings.ToLower(s) == "true":
 		patterns = []regex.Regex{matchAllRegex}
 	default:
 		splitted := strings.Split(s, "\n")
```
gandarez, line 636: will this split work since you already trimmed by `\n\r\t ` before?

pkg/project/project.go
alanhamlett: Comments are bad practice... the code should be the comment.

pkg/project/project.go
```diff
@@ -18,12 +18,14 @@ import (
 // nolint: gochecknoglobals
 var driveLetterRegex = regexp.MustCompile(`^[a-zA-Z]:\\$`)
 
-// WakaTimeProjectFile is the special file which if present should contain the project name and optional branch name
-// that will be used instead of the auto-detected project and branch names.
-const WakaTimeProjectFile = ".wakatime-project"
+const (
+	// WakaTimeProjectFile is the special file which if present should contain the project name and optional branch name
+	// that will be used instead of the auto-detected project and branch names.
+	WakaTimeProjectFile = ".wakatime-project"
```
gandarez: Don't need actually blank space here, let's try to stick to what we already have done for the other files.

gandarez (changes requested):
I think the default switch for `parseBoolOrRegexList` won't work (vide my comment). Can we also add tests for this func?

cmd/legacy/heartbeat/params.go
alanhamlett: Yes, before I only removed whitespace from start and end so it can still split by whitespace contained within the string.

gandarez: I would consider having tests for it.

pkg/project/project.go
alanhamlett: Checked and you're right, it makes no difference. I remember I was going to change `filepath.Join(startDir, "..")` to `filepath.Dir(startDir)` but using `Join` with `".."` avoids the whole problem.

cmd/legacy/heartbeat/params.go
alanhamlett: Added a test for it with b71f5c1d521ed994b65b7610efaa9d62d1282ed7.

gandarez (commented):
I think we don't need the new test `TestLoadParams_Filter_Exclude_Multiline` since it's only testing partially the returned value. I commited a suggestion [here](https://github.com/wakatime/wakatime-cli/commit/6d58bf6c4b0294141a96d5beee67b9a145fe7610) for testing and also a fix for windows new lines. Would you consider adding them and removing the test you added maybe?

alanhamlett:
> I think we don't need the new test `TestLoadParams_Filter_Exclude_Multiline` since it's only testing partially the returned value. I commited a suggestion [here](https://github.com/wakatime/wakatime-cli/commit/6d58bf6c4b0294141a96d5beee67b9a145fe7610) for testing and also a fix for windows new lines. Would you consider adding them and removing the test you added maybe?

Looks good, I applied it with a minor modification to support old Mac style (only `\r`).

gandarez (approved):
🚀

gandarez:
I think it needs some rebasing and squashing dangling commits.

PR #610, opened 2021-12-26
Send only 25 heartbeats per http request and save leftovers to offline db

### LINKED ISSUE #605: Too many heartbeats sent in one api request
Noticed in #600 an unrelated warning that means we're trying to send more than 25 heartbeats per HTTPS request to the API.

```
{"message":"Too many heartbeats", "caller":"/wakatime-cli/cmd/legacy/heartbeat/heartbeat.go:166","func":"SendHeartbeats","level":"warning","now":"2021-12-14T21:30:36-06:00","version":"v1.30.2"}
```

gandarez:
This PR limits to only 25 heartbeats at once sent to the the API and save leftovers to offline db. It also unifies **heartbeat.Params** (_from cmd/heartbeat package_) into **cmd/params** to avoid cycle import between the new package **cmd/offline** and **cmd/heartbeat**. Some changes were necessary to adapt the new command structures. In addition new unit and integration tests were added to cover the expected behaviour of limiting to 25 heartbeats per http request.

Fixes https://github.com/wakatime/wakatime-cli/issues/605


pkg/offline/offline.go
```diff
@@ -32,7 +26,10 @@ const (
 	maxRequeueAttempts = 3
 	// sendLimit is the maximum number of heartbeats, which will be sent at once
 	// to the WakaTime API.
-	sendLimit = 10
+	sendLimit = 24
```
alanhamlett: How is this different from `sendLimit` in `cmd/heartbeat/heartbeat.go`?

gandarez: Nothing, both have the same purpose. Just forgot to merge in a single one.

pkg/offline/offline_test.go
```diff
@@ -454,7 +454,7 @@ func TestSync_MultipleRequests(t *testing.T) {
 	dataGo, err := os.ReadFile("testdata/heartbeat_go.json")
 	require.NoError(t, err)
 
-	for i := 0; i < 11; i++ {
+	for i := 0; i < 26; i++ {
```
gandarez, line 457:
```suggestion
	for i := 0; i < 36; i++ {
```

cmd/offline/offline.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+
+			heartbeats = append(heartbeats, heartbeat.New(
+				h.Category,
+				h.CursorPosition,
+				h.Entity,
+				h.EntityType,
+				h.IsWrite,
+				h.Language,
+				"",
```
alanhamlett: Extra heartbeats can include alternate language too. Should this be `h.AlternateLanguage`?

alanhamlett (approved):

cmd/offline/offline.go
gandarez: Yep it should be.

alanhamlett (approved):
Awesome!!

PR #612, opened 2021-12-29
Add timezone data to http request

gandarez:
This PR adds timezone data from user's machine to http requests. The [go-olson-timezone](https://github.com/gandarez/go-olson-timezone) is a port from python tzlocal used in the legacy cli.


pkg/timezone/timezone_unix.go
```diff
@@ -0,0 +1,25 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to read symlink for localtime")
+	}
+
+	idx := strings.Index(tz, "zoneinfo/")
+	if idx == -1 {
+		return "", fmt.Errorf("'zoneinfo' not found at path for %s", runtime.GOOS)
+	}
+
+	return tz[idx+9:], nil
```
alanhamlett: This may result in `panic: runtime error: slice bounds out of range` so let's check the length of `tz` beforehand.

pkg/timezone/timezone_windows.go
```diff
@@ -0,0 +1,181 @@
...
+	// just work around it.
+	tzwin = strings.ReplaceAll(tzwin, "\x00", "")
+
+	if tz, ok := timezones[tzwin]; ok {
+		return tz, nil
+	}
+
+	return "", fmt.Errorf("windows timezone '%s' not found", tzwin)
+}
+
```
alanhamlett: This map looks the same as [tzlocal/windows_tz.py](https://github.com/regebro/tzlocal/blob/d5fb741969269c1cc0f4c1a87a66c10426863761/tzlocal/windows_tz.py) but that file contains many more zone mappings. Should we add those here?

pkg/timezone/timezone_windows.go
```diff
@@ -0,0 +1,181 @@
...
+var timezones = map[string]string{
```
alanhamlett: Tests failing because we need a different variable name instead of `timezones`.

pkg/timezone/timezone_unix.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"strings"
+)
+
+// Timezone returns the local timezone if found. If not, an error is returned.
+func Timezone() (string, error) {
+	tz, err := os.Readlink("/etc/localtime")
```
alanhamlett: According to [tzlocal/unix.py](https://github.com/regebro/tzlocal/blob/d5fb741969269c1cc0f4c1a87a66c10426863761/tzlocal/unix.py) it's slightly more complicated than just reading `/etc/localtime` depending on the flavor of Unix/Linux. Let's try to match the logic in tzinfo?

alanhamlett (changes requested):

cmd/api/api.go
```diff
@@ -32,6 +35,13 @@ func newClient(params paramscmd.API, opts ...api.Option) (*api.Client, error) {
 	opts = append(opts, api.WithTimeout(params.Timeout))
 	opts = append(opts, api.WithHostname(params.Hostname))
 
+	tz, err := timezone.Name()
+	if err != nil {
+		log.Debugf("failed to set up timezone option on api client: %s", err)
```
alanhamlett: it sounds like an error, how about `unable to detect local timezone`?

alanhamlett (approved):

alanhamlett:
Love the new `go-olson-timezone` module!

cmd/api/api.go
gandarez: changed to `failed to detect local timezone` tks :)

PR #615, opened 2022-01-02
Add project folder obfuscation

### LINKED ISSUE #591: Config option to only send file paths relative to project folder
### Discussed in https://github.com/wakatime/wakatime-cli/discussions/589

<div type='discussions-op-text'>

<sup>Originally posted by **kaushalyap** November 28, 2021</sup>
Currently it send absolute paths, I am aware of `hide_file_names option` As I think this obfuscate all file names, instead I want to obfuscate only the path to project, but still keep the filenames relative to the project directory ex: `{obfuscateted}/ProjectName/somefile.js`</div>

gandarez:
This PR adds project folder obfuscation to not send to API the root directory and only track the project's folder.

* Because we can't exactly know the root folder on map detection then the entire entity path will be considered. Like `"/usr/tmp/project/folder/entity.any"` will become `"/usr/tmp/project/folder/"`.
* For others we can assume the auto-detected filepath.

Closes https://github.com/wakatime/wakatime-cli/issues/591

### DIFF
--- a/cmd/params/params.go
+++ b/cmd/params/params.go
@@ -114,9 +114,11 @@ type (
 
 	// SanitizeParams params for heartbeat sanitization.
 	SanitizeParams struct {
-		HideBranchNames  []regex.Regex
-		HideFileNames    []regex.Regex
-		HideProjectNames []regex.Regex
+		HideBranchNames     []regex.Regex
+		HideFileNames       []regex.Regex
+		HideProjectFolder   bool
+		HideProjectNames    []regex.Regex
+		ProjectPathOverride string
 	}
 
 	// StatusBar contains status bar related parameters.
@@ -408,8 +410,6 @@ func loadFilterParams(v *viper.Viper) FilterParams {
 }
 
 func loadSanitizeParams(v *viper.Viper) (SanitizeParams, error) {
-	var params SanitizeParams
-
 	// hide branch names
 	hideBranchNamesStr, _ := vipertools.FirstNonEmptyString(
 		v,
@@ -428,8 +428,6 @@ func loadSanitizeParams(v *viper.Viper) (SanitizeParams, error) {
 		)
 	}
 
-	params.HideBranchNames = hideBranchNamesPatterns
-
 	// hide project names
 	hideProjectNamesStr, _ := vipertools.FirstNonEmptyString(
 		v,
@@ -447,8 +445,6 @@ func loadSanitizeParams(v *viper.Viper) (SanitizeParams, error) {
 			err,
 		)
 	}
-
-	params.HideProjectNames = hideProjectNamesPatterns
 
 	// hide file names
 	hideFileNamesStr, _ := vipertools.FirstNonEmptyString(
@@ -470,16 +466,16 @@ func loadSanitizeParams(v *viper.Viper) (SanitizeParams, error) {
 		)
 	}
 
-	params.HideFileNames = hideFileNamesPatterns
-
-	return params, nil
+	return SanitizeParams{
+		HideBranchNames:     hideBranchNamesPatterns,
+		HideFileNames:       hideFileNamesPatterns,
+		HideProjectFolder:   vipertools.FirstNonEmptyBool(v, "hide-project-folder", "settings.hide_project_folder"),
+		HideProjectNames:    hideProjectNamesPatterns,
+		ProjectPathOverride: vipertools.GetString(v, "project-folder"),
+	}, nil
 }
 
 func loadProjectParams(v *viper.Viper) (ProjectParams, error) {
-	if v == nil {
-		return ProjectParams{}, errors.New("viper instance unset")
-	}
-
 	disableSubmodule, err := parseBoolOrRegexList(vipertools.GetString(v, "git.submodules_disabled"))
 	if err != nil {
 		return ProjectParams{}, fmt.Errorf(
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -103,6 +103,12 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
 	flags.String("hide-file-names", "", "Obfuscate filenames. Will not send file names to api.")
 	flags.String("hide-filenames", "", "(deprecated) Obfuscate filenames. Will not send file names to api.")
 	flags.String("hidefilenames", "", "(deprecated) Obfuscate filenames. Will not send file names to api.")
+	flags.Bool(
+		"hide-project-folder",
+		false,
+		"When set, send the file's path relative to the project folder."+
+			" For ex: /User/me/projects/bar/src/file.ts is sent as src/file.ts so the server never sees the full path."+
+			" When the project folder cannot be detected, only the file name is sent. For ex: file.ts.")
 	flags.String(
 		"hide-project-names",
 		"",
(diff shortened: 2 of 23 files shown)


cmd/params/params.go
```diff
@@ -470,16 +466,16 @@ func loadSanitizeParams(v *viper.Viper) (SanitizeParams, error) {
...
+		HideBranchNames:     hideBranchNamesPatterns,
+		HideFileNames:       hideFileNamesPatterns,
+		HideProjectFolder:   vipertools.FirstNonEmptyBool(v, "relative-paths", "settings.relative_paths"),
+		HideProjectNames:    hideProjectNamesPatterns,
+		OverrideProjectPath: vipertools.GetString(v, "project-folder"),
+	}, nil
 }
 
 func loadProjectParams(v *viper.Viper) (ProjectParams, error) {
-	if v == nil {
```
alanhamlett, line 479: Just want to make sure this won't cause a panic if for some reason `v` is `null`?

cmd/root.go
```diff
@@ -150,18 +150,27 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
...
+		"(internal) Specify an offline queue file, which will be used instead of the default one.",
 	)
 	flags.String("plugin", "", "Optional text editor plugin name and version for User-Agent header.")
 	flags.String("project", "", "Override auto-detected project."+
 		" Use --alternate-project to supply a fallback project if one can't be auto-detected.")
+	flags.String(
+		"project-folder",
+		"",
+		"Optional workspace path. Override auto-detected project folder."+
+			" Useful only when --relative-paths is true.")
```
alanhamlett: Optional workspace path. Only used when `relative_paths=true` in the config file, or `--relative-paths` flag is present.

cmd/params/params.go
gandarez: Null check is done at the `Load` func.

cmd/root.go
```diff
@@ -150,18 +150,27 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
...
 		"proxy",
 		"",
 		"Optional proxy configuration. Supports HTTPS SOCKS and NTLM proxies."+
 			" For example: 'https://user:pass@host:port' or 'socks5://user:pass@host:port'"+
 			" or 'domain\\user:pass'",
 	)
+	flags.Bool(
+		"relative-paths",
+		false,
+		"When set, left-trim the absolute path removing 'workspace-path'.")
```
alanhamlett: When set, send the file's path relative to the project folder. For ex: /User/me/projects/bar/src/file.ts is sent as src/file.ts so the server never sees the full path. When the project folder cannot be detected, only the file name is sent. For ex: file.ts.

alanhamlett: Let's rename this to match the ini setting: `--hide-project-folder`

cmd/params/params.go
```diff
@@ -470,16 +466,16 @@ func loadSanitizeParams(v *viper.Viper) (SanitizeParams, error) {
 		)
 	}
 
-	params.HideFileNames = hideFileNamesPatterns
-
-	return params, nil
+	return SanitizeParams{
+		HideBranchNames:     hideBranchNamesPatterns,
+		HideFileNames:       hideFileNamesPatterns,
+		HideProjectFolder:   vipertools.FirstNonEmptyBool(v, "relative-paths", "settings.relative_paths"),
```
alanhamlett: Should we rename this to `settings.hide_project_folder`?

pkg/heartbeat/sanitize.go
```diff
@@ -64,6 +68,15 @@ func Sanitize(h Heartbeat, config SanitizeConfig) Heartbeat {
 		h.Branch = nil
 	}
 
+	if h.EntityType == FileType && config.HideProjectFolder {
+		// normalize project path to look like a directory
+		if !strings.HasSuffix(h.ProjectPath, string(os.PathSeparator)) {
+			h.ProjectPath += string(os.PathSeparator)
+		}
+
+		h.Entity = strings.TrimPrefix(h.Entity, h.ProjectPath)
```
alanhamlett: We should actually trim both `--project-folder` and the auto detected project folder from `h.Entity` because someone might open a file from another project. The `--project-folder` wouldn't match, but the auto-detected folder would.

alanhamlett (changes requested):
Sorry, my instructions were flawed. Left some comments.

cmd/root.go
gandarez: sure

cmd/params/params.go
gandarez: makes more sense to be explicitly defined.

alanhamlett: So we leave it unchanged (`relative_paths`)?

gandarez: `hide_project_folder` is more meanful IMO

alanhamlett: Yep, I agree. So `--hide-project-folder` and `settings.hide_project_folder = true`.

pkg/project/file.go
```diff
@@ -27,11 +28,12 @@ func (f File) Detect() (Result, bool, error) {
...
 	lines, err := readFile(fp, 2)
 	if err != nil {
-		return Result{}, false,
-			Err(fmt.Sprintf("error reading file: %s", err))
+		return Result{}, false, Err(fmt.Sprintf("error reading file: %s", err))
 	}
 
-	result := Result{}
+	result := Result{
+		Folder: filepath.Join(fp, ".."),
```
alanhamlett: It's clearer to use `Folder: filepath.Dir(fp),` unless that could cause problems from sometimes returning `"."` instead of empty string `""`?

gandarez: makes more sense to use `Dir(..)`, will change in the next fixup.

gandarez:
All fixes are up

pkg/heartbeat/sanitize.go
```diff
@@ -64,6 +67,22 @@ func Sanitize(h Heartbeat, config SanitizeConfig) Heartbeat {
 		h.Branch = nil
 	}
 
+	if h.EntityType == FileType && config.HideProjectFolder {
+		// normalize project path to look like a directory
+		if h.ProjectPath != "" && !strings.HasSuffix(h.ProjectPath, "/") {
+			h.ProjectPath += "/"
```
alanhamlett, line 84: Let's also replace `\` with `/` in `h.ProjectPath` using `windows.FormatFilePath`.

gandarez: Formatting auto-detected file path into project detection where it's being set.

pkg/project/file.go
```diff
@@ -27,11 +28,12 @@ func (f File) Detect() (Result, bool, error) {
...
 	lines, err := readFile(fp, 2)
 	if err != nil {
-		return Result{}, false,
-			Err(fmt.Sprintf("error reading file: %s", err))
+		return Result{}, false, Err(fmt.Sprintf("error reading file: %s", err))
 	}
 
-	result := Result{}
+	result := Result{
+		Folder: filepath.Dir(fp),
```
alanhamlett, line 35: I don't think we're running `windows.FormatFilePath` on this before using in `hideProjectFolder`. Let's just call `windows.FormatFilePath` inside `hideProjectFolder` to be sure no matter where the project folder comes from that it's normalized?

gandarez: We are. Take a look at `project.go` line 100. Project detection runs before sanitization.

alanhamlett: Oh, got it. It runs through `windows.FormatFilePath` no matter which type of project detection is used.

alanhamlett: Also, we're formatting the `--project-folder` at https://github.com/wakatime/wakatime-cli/pull/615/files#diff-d540dc518660d926557564ec01ffcf0beb6ef055d37b97413412597667850059R90 so both sources of project folder, and Entity are normalized to forward slash. 👍

alanhamlett (approved):
Ready to merge!

PR #618, opened 2022-01-04
Fix windows network mount regex

gandarez:
This PR fixes windows network mount regex to also support ip addresses. For ex: `\\192.168.1.1\some\project\folder\entity.any`. 

In addition it fixes `parseNetUseOutput` func to ignore statuses different of `OK` or string empty (_might happen and can be considered "OK" as well_). And also skip getting absolute and real path for network drives.

Example of `net use`
```
New connections will be remembered.


Status       Local     Remote                    Network

-------------------------------------------------------------------------------
Unavailable  I:        \\192.168.0.10\Documentos Microsoft Windows Network
OK           J:        \\192.168.68.110\Desenvolvimento
                                                Microsoft Windows Network
The command completed successfully.
```

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -14,7 +14,7 @@ import (
 var (
 	backslashReplaceRgx    = regexp.MustCompile(`[\\/]+`)
 	windowsDriveRgx        = regexp.MustCompile("^[a-z]:/")
-	windowsNetworkMountRgx = regexp.MustCompile(`(?i)^\\\\[a-z]+`)
+	windowsNetworkMountRgx = regexp.MustCompile(`(?i)^\\\\([a-z]|((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))+`)
 )
 
 // FormatFilePath formats a windows filepath by converting backslash to
@@ -35,6 +35,11 @@ func FormatFilePath(fp string) (string, error) {
 	}
 
 	return fp, nil
+}
+
+// IsWindowsNetworkMount returns true if filepath is windows network path.
+func IsWindowsNetworkMount(fp string) bool {
+	return windowsNetworkMountRgx.MatchString(fp)
 }
 
 // commander is an interface for exec.Command function.
@@ -127,7 +132,7 @@ func parseNetUseOutput(text string) (remoteDrives, error) {
 	drives := make(remoteDrives)
 
 	for _, line := range lines[1 : len(lines)-1] {
-		if len(strings.TrimSpace(line)) == 0 {
+		if len(strings.TrimSpace(line)) == 0 || strings.ContainsAny(line, "---") {
 			continue
 		}
 
@@ -137,6 +142,11 @@ func parseNetUseOutput(text string) (remoteDrives, error) {
 				return nil, fmt.Errorf("failed to parse columns from 'net use' output: %s", err)
 			}
 
+			continue
+		}
+
+		status := strings.TrimSpace(line[:cols.Local.Start])
+		if status != "" && strings.ToLower(status) != "ok" {
 			continue
 		}
 
@@ -216,7 +226,7 @@ func parseNetUseColumns(line string) (netUseColumns, error) {
 	}
 
 	if cols.Local.Empty() {
-		return netUseColumns{}, errors.New("failed to parse local columns")
+		return netUseColumns{}, errors.New("failed to parse local column")
 	} else if cols.Remote.Empty() {
 		return netUseColumns{}, errors.New("failed to parse remote column")
 	}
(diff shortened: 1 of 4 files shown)


pkg/windows/windows.go
```diff
@@ -14,7 +14,7 @@ import (
 var (
 	backslashReplaceRgx    = regexp.MustCompile(`[\\/]+`)
 	windowsDriveRgx        = regexp.MustCompile("^[a-z]:/")
-	windowsNetworkMountRgx = regexp.MustCompile(`(?i)^\\\\[a-z]+`)
+	windowsNetworkMountRgx = regexp.MustCompile(`(?i)^\\\\([a-z]|((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))+`)
```
alanhamlett: Can this be simplified to `(?i)^\\\\((\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})|(\w+))\\`?

pkg/windows/windows.go
```diff
@@ -140,6 +145,11 @@ func parseNetUseOutput(text string) (remoteDrives, error) {
 			continue
 		}
 
+		status := strings.TrimSpace(line[:cols.Local.Start])
+		if status != "" && strings.ToLower(status) != "ok" {
```
alanhamlett: Should we still parse network drives which aren't `ok`? What if a drive is showing temporarily as `unavailable`, we would still want to detect the path as a remote drive.

gandarez: the `unavailable` status is often shown when the mapped network drive has an invalid ip address. I demonstrate that in the example.

pkg/windows/windows.go
alanhamlett: How about ipv6? https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses

pkg/windows/windows.go
alanhamlett: It's ok to still parse though, because the `windowsNetworkMountRgx` should prevent it from being used? I think it's better to parse all listed remote drives in case we have unexpected cases where a valid drive isn't listed as `ok`.

gandarez: Valid drives are never listed as `ok` if them aren't `ok` and accessible. What I did is to consider empty and `ok` statuses to avoid unavailable ones.

pkg/windows/windows.go
gandarez: `((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)` is the offical regex for ip address so that’s why I used it.

it accepts:
```
127.0.0.1
192.168.1.1
192.168.1.255
255.255.255.255
0.0.0.0
```
and rejects:
```
30.168.1.255.1
127.1
192.168.1.256
-1.2.3.4
1.1.1.1.
3...3
```

gandarez: Would you like to parse both v4 and v6?

pkg/windows/windows.go
alanhamlett: Would it cause problems to parse it anyway?

gandarez: It doesn't cause problems when parsing but we're taking the risk of using unavailable mappings as valid ones.

alanhamlett: Can there be duplicate entries for the same drive letter? If not, then it shouldn't matter if we parse unavailable ones because they just won't be used since the entity file path would only be using drives that are readable in the IDE.

pkg/windows/windows.go
```diff
@@ -140,6 +145,11 @@ func parseNetUseOutput(text string) (remoteDrives, error) {
...
+			continue
+		}
+
 		local := line[cols.Local.Start : cols.Local.Start+cols.Remote.Width]
 		local = strings.ToUpper(strings.TrimSpace(local))
 		letter := strings.Split(local, ":")[0][0]
```
alanhamlett, line 174: If the drive letter is empty, this will result in a panic:
`panic: runtime error: index out of range [0] with length 0`

We need to check the length of `strings.Split(local, ":")[0]` before trying to take the first char.

gandarez: It would never happen because a drive letter is always linked to a remote path and vice-versa?

gandarez:
@alanhamlett added support fo ipv6 as well removed validation of status.

pkg/windows/windows.go
alanhamlett: In searching online i'm seeing examples where the drive letter is empty. Doesn't hurt to guard against it.

gandarez: it's already safe. Take a look at the [new test added](https://github.com/wakatime/wakatime-cli/pull/618/commits/1f0148e9fd63d681137e37aee9a3c54cbe892048#diff-d344b69e095a2125ed47fb159fbd42d2ef90c351badecd786c1514d6399b51ffR27).

alanhamlett (approved):

PR #620, opened 2022-01-06
Remove duplicated validation for file exists

gandarez:
This PR removes duplicated validation for file exists. The `filter.WithFiltering()` already takes care of filtering in case of missing file.

By removing the code it also fixed a bug when if any of extra heartbeat doesn't exist it returns `null` bypassing all parsed heartbeats. 
```
if h.EntityType == heartbeat.FileType && !isFile(h.Entity) {
    log.Warnf("file '%s' does not exist. ignoring this extra heartbeat", h.Entity)
    return nil
}
```
This PR is also a preparation for the next one.

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -3,7 +3,6 @@ package heartbeat
 import (
 	"errors"
 	"fmt"
-	"os"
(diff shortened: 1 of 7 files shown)


alanhamlett (approved):
Looks great!

PR #623, opened 2022-01-06
Add support for reading files over ssh or sftp

### LINKED ISSUE #426: Support reading remote files via ssh, sftp, and any other network drives

gandarez:
This PR adds support for reading files over ssh or sftp. It detects an entity as a remote file and tries to download the file to a temporary folder. The `LocalFile` field is assigned with the temporary full path and used afterhand in the pipeline. 

Closes https://github.com/wakatime/wakatime-cli/issues/426


pkg/remote/remote.go
```diff
@@ -0,0 +1,237 @@
...
+	defer srcFile.Close()
+
+	dstFile, err := os.Create(localFile)
+	if err != nil {
+		return fmt.Errorf("failed to open local file: %s", err)
+	}
+
+	defer dstFile.Close()
+
+	_, err = io.Copy(dstFile, srcFile)
```
alanhamlett: Let's use [io.CopyN](https://pkg.go.dev/io#CopyN) like:
`_, err = io.Copy(dstFile, srcFile, 512000)`

alanhamlett:
This is awesome!

pkg/language/chroma.go
```diff
@@ -42,6 +42,9 @@ import (
 	"github.com/danwakefield/fnmatch"
 )
 
+// Max file size supporting reading from file. Default is 512Kb.
+const maxFileSize = 512000
```
alanhamlett, line 46: Since this constant is repeated in a few places should we move it into a common place and import it everywhere?

pkg/heartbeat/heartbeat.go
```diff
@@ -19,6 +19,7 @@ type Heartbeat struct {
 	CursorPosition      *int       `json:"cursorpos"`
 	Dependencies        []string   `json:"dependencies"`
 	Entity              string     `json:"entity"`
+	EntityRaw           string     `json:"-"`
```
alanhamlett, line 22: I don't think we're populating this with a value anywhere yet?

gandarez: At `remote.WithDetection()`?

pkg/language/chroma.go
gandarez: in Go is not recommended to create packages like utils, helpers as for other programming languages. As both values can change time by time I suggest to keep them as is.

pkg/heartbeat/heartbeat.go
alanhamlett: Somehow didn't see it before.

alanhamlett (approved):

PR #629, opened 2022-01-25
Fix filtering unknown project

### LINKED ISSUE #627: Do not ignore the projects whose name is unknown, but available with vcs.
**Is your feature request related to a problem? Please describe.**

I am always see an 'Unknown Project' in my dashboard. So I want to enable the setting `exclude_unknown_project = true`, but then there are no project in my dashboard, and the log says all the projects (detected by vcs) have been ignored.

**Describe the solution you'd like to see**

Do not ignore the projects whose name is unknown, but available with vcs.

**Describe alternatives you've considered**

In another word, I don't want to see the project named `Unknown Project` on the dashboard, it takes up most of my work time, and I want to ignore it.

**Additional context**

Now, the project without name will be ignored.

https://github.com/wakatime/wakatime-cli/blob/7546bb237cf03ea00fd9062bd48017f3bdfacb43/pkg/filter/filter.go#L59-L61

When I perform some edit, the log says the heartbeat is ignored.

It seems that the projectname detected by vcs is  `alternate`.

Log:

```
{"caller":"/home/runner/work/wakatime-cli/wakatime-cli/cmd/heartbeat/heartbeat.go:85","func":"SendHeartbeats","level":"debug","message":"params: api params: (api key: '3760...', api url: 'https://api.wakatime.com/api/v1', backoff at: '', backoff retries: 0, hostname: '******', plugin: 'vscode/1.63.0 vscode-wakatime/18.0.4', timeout: 2m0s, disable ssl verify: false, proxy url: '', ssl cert filepath: ''), heartbeat params: (category: 'coding', cursor position: '1', entity: '/home/cubercsl/Project/wakatime-cli/pkg/heartbeat/heartbeat.go', entity type: 'file', num extra heartbeats: 0, is write: false, language: '', line number: '1', lines in file: '187', time: 1643030384.73025, filter params: (exclude: '[COMMIT_EDITMSG$ PULLREQ_EDITMSG$ MERGE_MSG$ TAG_EDITMSG$]', exclude unknown project: true, include: '[]', include only with project file: false), project params: (alternate: 'wakatime-cli', disable submodule: '[]', map patterns: '[]', override: ''), sanitize params: (hide branch names: '[]', hide project folder: false, hide file names: '[]', hide project names: '[]', project path override: '')), offline params: (disabled: false, queue file: '', num sync max: 1000), status bar params: (hide categories: false)","now":"2022-01-24T21:19:44+08:00","version":"v1.35.0"}
{"caller":"/home/runner/work/wakatime-cli/wakatime-cli/pkg/heartbeat/format.go:25","func":"1","level":"debug","message":"execute heartbeat filepath formatting","now":"2022-01-24T21:19:44+08:00","version":"v1.35.0"}
{"caller":"/home/runner/work/wakatime-cli/wakatime-cli/pkg/filter/filter.go:29","func":"1","level":"debug","message":"execute heartbeat filtering","now":"2022-01-24T21:19:44+08:00","version":"v1.35.0"}
{"caller":"/home/runner/work/wakatime-cli/wakatime-cli/pkg/filter/filter.go:36","func":"1","level":"error","message":"skipping because of unknown project","now":"2022-01-24T21:19:44+08:00","version":"v1.35.0"}
```

gandarez:
This PR fixes filtering unknown projects. Currently filtering for undetected project is being done before actually running vcs detection. It moves the logic for filtering from `filter` package to `project` and runs immediately after project detection.

fixes #627


alanhamlett (approved):

pkg/project/filter.go
```diff
@@ -0,0 +1,54 @@
...
+	return func(next heartbeat.Handle) heartbeat.Handle {
+		return func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
+			log.Debugln("execute project filtering")
+
+			var filtered []heartbeat.Heartbeat
+
+			for _, h := range hh {
+				err := Filter(h, config)
+				if err != nil {
+					log.Debugln("skipping because of unknown project")
```
dron22: The name of the function (`Filter`) is very generic. Thus I would not assume at this point, that every error happened due to an empty project. How about using the message from `err.Error()`? In the current implementation this will result in exactly the same log message.

pkg/project/project.go
```diff
@@ -76,28 +76,29 @@ func WithDetection(c Config) heartbeat.HandleOption {
...
-					revControlResult := DetectWithRevControl(h.Entity, c.SubmodulePatterns, c.ShouldObfuscateProject)
+					revControlResult := DetectWithRevControl(h.Entity, config.SubmodulePatterns)
 
 					result.Branch = firstNonEmptyString(result.Branch, revControlResult.Branch)
 					result.Folder = firstNonEmptyString(result.Folder, revControlResult.Folder)
-					result.Project = firstNonEmptyString(result.Project, revControlResult.Project)
 
-					if result.Project == "" {
-						result.Project = setProjectName(h.ProjectAlternate, c.ShouldObfuscateProject, revControlResult.Folder)
+					if config.ShouldObfuscateProject && revControlResult.Project != "" && result.Project == "" {
```
dron22, line 93: I guess here it would be helpful to explain in a comment, in which cases we obfuscate and in which not. Just to provide some more info on the underlying logic.

dron22 (approved):
LGTM :+1: Just 2 small remarks

pkg/project/filter.go
gandarez: That makes sense and I didn't see that `log.Debugln()` was printing the same error message coming from the function. Nice catch!

pkg/project/project.go
gandarez: yep sounds good.

PR #640, opened 2022-02-03
Improve loading params

gandarez:
This PR originally was intended to remove `QueueFilepathWithErr()` but other improvements were necessary and all are described and explained below. In a nutshell them are small ones but many lines were needed to changes specially at tests

* Both `QueueFilepath()` and `QueueFilepathWithErr()` have the same logic but one is not returning error. To avoid duplicated code `QueueFilepathWithErr()` has been removed in order to have only `QueueFilepath() (string, error)`.
* The default folder returned when `ini.WakaHomeDir()` fails was always a join of "empty" folder with ".wakatime.bdb" so it has been fixed to simply return `.wakatime.bdb`.
* The params config struct was removed in order to simplify `params.Load()` function. Taking that approach few functions were made public to be called individually and make the error handling better. Functions are `LoadHeartbeatParams`, `LoadAPIParams`, `LoadOfflineParams` and `LoadStausBarParams`. Also all unit tests of `params_test.go` were calling the `Load()` function and were replaced accordingly.
* The flags `--today`, `--today-goal`, `--sync-offline-activity` and `--offline-count` have also been changed to better handle error when parsing config file.
* The `cmd.offline.go` dropped the usage of `params.Load()` in order to call them individually and again to get better error handling. Also a bug was fixed in a rare condition where `heartbeats` could be not nil and any heartbeat param was loaded where them are required at `initHandleOptions()`. 
* The `cmd.heartbeat.go` still calls `params.Load()` since it needs all of them loaded correctly as it's designed.
* At the `cmd.run.go` I introduced a call to `SetupLogging()` when `parseConfigFiles()` fails. It basically sets the log to output to file instead of the std out which at this time isn't the wakatime log file .
* Changed `sendDiagnostics()` to return an error and properly handle it at the caller as a warning.
* The `ini` package has been fixed to always return an empty string when any error occurs. It also logs as warning and goes to the next step. So it always try to get the home directory unless all steps fail.

### DIFF
--- a/cmd/offline/offline.go
+++ b/cmd/offline/offline.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/wakatime/wakatime-cli/cmd/params"
+	paramscmd "github.com/wakatime/wakatime-cli/cmd/params"
 	"github.com/wakatime/wakatime-cli/pkg/deps"
 	"github.com/wakatime/wakatime-cli/pkg/filestats"
 	"github.com/wakatime/wakatime-cli/pkg/filter"
@@ -21,15 +21,8 @@ import (
 // SaveHeartbeats saves heartbeats to the offline db, when we haven't
 // tried sending them to the API. If we tried sending to API already,
 // to the API. Used when we have heartbeats unsent to API.
-func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat) error {
-	config := params.Config{
-		ForSavingOffline: true,
-	}
-	if heartbeats == nil {
-		config.HeartbeatRequired = true
-	}
-
-	params, err := params.Load(v, config)
+func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat, queueFilepath string) error {
+	params, err := loadParams(v, heartbeats == nil)
 	if err != nil {
 		return fmt.Errorf("failed to load command parameters: %w", err)
 	}
@@ -50,7 +43,6 @@ func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat) error {
 
 	handleOpts := initHandleOptions(params)
 
-	queueFilepath := offline.QueueFilepath()
 	if params.Offline.QueueFile != "" {
 		queueFilepath = params.Offline.QueueFile
 	}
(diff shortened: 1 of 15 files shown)


cmd/offline/offline.go
```diff
@@ -70,6 +63,29 @@ func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat) error {
 	return nil
 }
 
+func loadParams(v *viper.Viper) (paramscmd.Params, error) {
+	paramHeartbeat, err := paramscmd.LoadHeartbeatParams(v)
+	if err != nil {
+		return paramscmd.Params{}, fmt.Errorf("failed to load heartbeat parameters: %s", err)
+	}
+
+	paramAPI, err := paramscmd.LoadAPIParams(v)
```
alanhamlett, line 66: We don't need to load api params when saving heartbeats to the offline db.

cmd/offline/offline.go
```diff
@@ -70,6 +63,29 @@ func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat) error {
 	return nil
 }
 
+func loadParams(v *viper.Viper) (paramscmd.Params, error) {
+	paramHeartbeat, err := paramscmd.LoadHeartbeatParams(v)
```
alanhamlett: When only saving extra heartbeats over the max, we don't want to load heartbeat params. That's the only case where we only want to save the given heartbeats without saving the param heartbeat.

cmd/offline/offline.go
gandarez: It's required for `setLogFields()` [here](https://github.com/wakatime/wakatime-cli/blob/misc/improve-load-params/cmd/offline/offline.go#L141) and `buildHeartbeats()` [here](https://github.com/wakatime/wakatime-cli/blob/misc/improve-load-params/cmd/offline/offline.go#L93).

alanhamlett: But we don't want the param heartbeat when only saving surplus extra heartbeats here:
https://github.com/wakatime/wakatime-cli/blob/f008c674afa6615bbaacacfb0c50d2401da74eb5/cmd/heartbeat/heartbeat.go#L97

alanhamlett: Oh, f008c674afa6615bbaacacfb0c50d2401da74eb5 fixes it 👍

alanhamlett (approved):

PR #649, opened 2022-02-10
Fix api key being hidden incorrectly

gandarez:
This PR fixes api key being hidden incorrectly and instead of last 4 chars it's considering the whole string but the first 4. I also replaced "..." with "<hidden>", what do you think?

```
<hidden>bb3c
```

### DIFF
--- a/cmd/params/params.go
+++ b/cmd/params/params.go
@@ -749,7 +749,7 @@ func (p API) String() string {
 	apiKey := p.Key
 	if len(apiKey) > 4 {
 		// only show last 4 chars of api key in logs
-		apiKey = "..." + apiKey[:len(apiKey)-4]
+		apiKey = fmt.Sprintf("<hidden>%s", apiKey[len(apiKey)-4:])
 	}
 
 	return fmt.Sprintf(


dron22 (commented):
My original intention was to be able to detect if the correct API key was used. Not by comparing a big portion of the key, but just by showing 4 characters to rule out, that something very different was used as api key.

In general, the more we show of the API key, the less "secure". So, what exactly is the reason for showing more characters of the API key? What do we gain here?

gandarez:
Last change incorrectly modified the way api key was hidden. The intention was to safely show the last 4 digits only when the lenght of the api key is greater than 4. This PR fixes it and instead skipping the first 4 it'll show the last 4.

cmd/params/params.go
```diff
@@ -749,7 +749,7 @@ func (p API) String() string {
 	apiKey := p.Key
 	if len(apiKey) > 4 {
 		// only show last 4 chars of api key in logs
-		apiKey = "..." + apiKey[:len(apiKey)-4]
```
alanhamlett, line 752: Why change `apiKey[:len(apiKey)-4]` to `apiKey[len(apiKey)-4:]`? They both return the last 4 characters of the string. Is one better than the other?

alanhamlett:
Using `...` is shorter than `<hidden>`, but I really don't have a preference between the two. The IDE extensions all use `XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX` so the log would look like `XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX1234`. That's ok for IDE logs because they usually get cleared when the IDE restarts, but for `~/.wakatime.log` that would be too long and make logs harder to read.

cmd/params/params.go
gandarez: https://go.dev/play/p/Xc_dRZwabyQ

alanhamlett: Oops! Thanks for catching that!

alanhamlett (approved):

Issue #654, opened 2022-02-17
Malware detected in wakatime-cli for Windows


markvandenbrink:
**Actual behavior (what went wrong)**:
According to Windows Defender, the latest version of the `wakatime-cli` contains malware (PUA:Win32/Caypnamer.A!ml). I first noticed this using the Wakatime extension in Visual Studio Code. When I directly download the latest release of the windows version from this repo (https://github.com/wakatime/wakatime-cli/releases/download/v1.37.0/wakatime-cli-windows-amd64.zip), the ZIP file gets immediately marked/blocked. I hope this is a false positive. Otherwise, this could be a serious problem...

**Environment**:
Windows 11

**Logs**:
![image](https://user-images.githubusercontent.com/7869636/154517722-489d251b-c7ea-4442-a5ec-e4bc6dfaccfd.png)

![image](https://user-images.githubusercontent.com/7869636/154518550-e6571b3f-ace0-433a-9279-627b58b20cf5.png)

arturohernandez10:
Same thing happened to me. I believe it's just suspecting wakatime to be a trojan. As I use VSCODE, the command prompt appears asking me to enter my API code. Essentially the HTTP call gets blocked. Just to be sure. I'll let Microsoft Defender block it for now,

alanhamlett:
This [has happened before](wakatime/wakatime-cli#693), and usually is fixed in a few days after we submit a request to MS. I'll let you know when we receive a reply.

alanhamlett:
We've asked Microsoft to unblock it, will update when we receive more info.


![Screen Shot 2022-02-17 at 10 53 09 AM](https://user-images.githubusercontent.com/522344/154551509-ec94c762-dbfe-4a69-906c-7b241f2d56ad.png)

alanhamlett:
Should be fixed now, but you may need to update your malware definitions for it to take affect:

1. Open command prompt as administrator and change directory to c:\Program Files\Windows Defender
2. Run “MpCmdRun.exe -removedefinitions -dynamicsignatures”
3. Run "MpCmdRun.exe -SignatureUpdate"

----

<img width="929" alt="Screen Shot 2022-02-17 at 1 41 43 PM" src="https://user-images.githubusercontent.com/522344/154575691-fd587ee4-6890-47a4-bc92-b9f073a8f8c3.png">

aggietallboy:
[...]

PR #656, opened 2022-02-22
Add threat scan for windows binary

gandarez:
This PR adds threat scanner for windows binary to avoid Windows Defender blocking it in some workstations when fully enabled. If threat scan fails we should manually upload the binary to https://www.microsoft.com/en-us/wdsi/filesubmission, wait for approval and then restart the workflow.


alanhamlett:
Awesome! This will prevent issues like #654 from happening in the future.

alanhamlett (approved):

PR #679, opened 2022-04-01
Fix sanitization for offline heartbeats

### LINKED ISSUE #676: Syncs after error leak sensitive data
If the `hidefilenames` attribute is set to true, all files should be renamed to HIDDEN._extension_

Something seems to go wrong with the file sanitization when the filter facilities (pkg/filter/filter.go) error out, which leads to the caching of unsanitized heartbeats in the `$HOME/wakatime.bdb` file, which I presume are then queued to be resend later on.

**Environment**:
Arch Linux x86

**Logs**:
Note to the logs: I'm running my own backend on https://rb4/wakapi/api/users/current/heartbeats.bulk which is available only internally in my network, but the same issues happens with the official backend.
The following error 
```
{"caller":"github.com/wakatime/wakatime-cli/pkg/heartbeat/format.go:53","func":"Format","level":"warning","message":"failed to resolve real path for \"/home/joerg/dev/a/jabcoin/NvimTree_1\": lstat /home/joerg/dev/a/jabcoin/NvimTree_1: no such file or directory","now":"2022-03-30T22:05:24+02:00","version":"1.35.4"}
```
is triggered by the neovim plugin nvimtree when the cursor is being moved around in the tree view, see the left side [here](https://raw.githubusercontent.com/kyazdani42/nvim-tree.lua/master/.github/screenshot.png).

Also notice that roughly midway through the logs the heartbeats are shown properly sanitized (line 143), but then towards the end they are listed again but completely unsanitizied (lines 313, 316).

https://gist.github.com/jabra11/3f11f780e06149b21baf878d53e722f6

gandarez:
This PR fixes heartbeats not sanitizeds in an specific situation causing sanitize params not being loaded. It also fixed:

1. Don't  need to return error for `offline.WithQueue()`.
2. Don't try to save heartbeats to offline db after `params.Load(v)` has failed since it needs params to be loaded and might fail.

To cover this scenario the following test was added:

* TestSendHeartbeats_ExtraHeartbeats_Sanitize

Fixes https://github.com/wakatime/wakatime-cli/issues/676

### DIFF
--- a/cmd/offline/offline.go
+++ b/cmd/offline/offline.go
@@ -22,7 +22,7 @@ import (
 // tried sending them to the API. If we tried sending to API already,
 // to the API. Used when we have heartbeats unsent to API.
 func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat, queueFilepath string) error {
-	params, err := loadParams(v, heartbeats == nil)
+	params, err := loadParams(v)
 	if err != nil {
 		return fmt.Errorf("failed to load command parameters: %w", err)
 	}
@@ -47,12 +47,7 @@ func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat, queueFilep
 		queueFilepath = params.Offline.QueueFile
 	}
 
-	offlineHandleOpt, err := offline.WithQueue(queueFilepath)
-	if err != nil {
-		return fmt.Errorf("failed saving heartbeats because unable to init offline queue: %w", err)
-	}
-
-	handleOpts = append(handleOpts, offlineHandleOpt)
+	handleOpts = append(handleOpts, offline.WithQueue(queueFilepath))
 
 	sender := offline.Sender{}
 	handle := heartbeat.NewHandle(&sender, handleOpts...)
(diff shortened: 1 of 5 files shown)


cmd/offline/offline.go
```diff
@@ -22,7 +22,7 @@ import (
 // tried sending them to the API. If we tried sending to API already,
 // to the API. Used when we have heartbeats unsent to API.
 func SaveHeartbeats(v *viper.Viper, heartbeats []heartbeat.Heartbeat, queueFilepath string) error {
-	params, err := loadParams(v, heartbeats == nil)
```
alanhamlett, line 25: Don't we want to skip re-loading the param heartbeat when only loading params for saving extra heartbeats? Otherwise we'll accidentally save the param heartbeat to offline db when we only wanted to save the extra heartbeats over the max.

https://github.com/wakatime/wakatime-cli/pull/640#discussion_r799632901

gandarez: We have a guard few lines below [here](https://github.com/wakatime/wakatime-cli/blob/bugfix/offline-sanitization/cmd/offline/offline.go#L38). That way we ensure to build heartbeats (_including the main one_) only when it comes null. We need to always load `HeartbeatParam` because all sanitization rules are in it.

alanhamlett (approved):

PR #684, opened 2022-04-07
Allow unsaved entity to be sent to the API

gandarez:
This PR allows any unsaved/empty entity/extra heartbeats to be sent to the API by setting the flag `is-unsaved-entity` to **true**. The following validations will be skipped:
* formatting
* remote detection
* dependency detection

Those ones will always run for every single heartbeat:
* entity filtering (with exceptions)
* entity modifier
* language detection
* project detection (with exceptions)
* project filtering
* sanitization

Feature to solve [this issue](https://github.com/wakatime/vscode-wakatime/issues/266).


cmd/heartbeat/heartbeat.go
```diff
@@ -193,42 +193,60 @@ func buildHeartbeats(params paramscmd.Params) []heartbeat.Heartbeat {
...
+		opts = append(opts,
+			remote.WithDetection(),
+			filestats.WithDetection(filestats.Config{
+				LinesInFile: params.Heartbeat.LinesInFile,
+			}),
+			language.WithDetection(),
+			deps.WithDetection(deps.Config{
+				FilePatterns: params.Heartbeat.Sanitize.HideFileNames,
+			}),
+			project.WithDetection(project.Config{
```
alanhamlett: Skipping project detection means the value from `--project` and `--alternate-project` aren't used?

cmd/root.go
```diff
@@ -35,6 +35,7 @@ func NewRootCMD() *cobra.Command {
 
 func setFlags(cmd *cobra.Command, v *viper.Viper) {
 	flags := cmd.Flags()
+	flags.Bool("allow-unsaved-entity", false, "Skip checking for file existing and any detection on the entity.")
```
alanhamlett: Let's use description:

`Normally files that don't exist on disk are skipped and not tracked. When this option is present, the file will be tracked even if it doesn't exist.`

alanhamlett (changes requested):


cmd/heartbeat/heartbeat.go
gandarez: Language and project detection can run under circunstances. I'm fixing it in this PR.

alanhamlett (changes requested):
I just realized, we need `config.AllowUnsavedEntity` tied to the heartbeat, because we don't want to prevent extra heartbeats that are real files to skip project and language detection.

gandarez:
> I just realized, we need `config.AllowUnsavedEntity` tied to the heartbeat, because we don't want to prevent extra heartbeats that are real files to skip project and language detection.

Do you mean a new attribute for extra heartbeats to individually control each one or only the root param is enough?

alanhamlett (approved):

PR #696, opened 2022-04-13
Add project_api_key setting parsing

### LINKED ISSUE #694: Use different api key per project folder
A new `~/.wakatime.cfg` INI section `[project_api_key]`.

Similar to the `[projectmap]` section, each key is a regex pattern matching the `entity` path. The value of each key is the api key to use when working in that project/folder. For ex:

```ini
[settings]
api_key = 1234
[project_api_key]
/projects/work/ = 5678
```

Normally, api key `1234` is used except when `entity` contains/matches regex `/projects/work/` and then api key `5678` is used.

gandarez:
This PR adds a new section `project_api_key` to config file to allow using different api key per project path. It also fixed reading `projectmap` by manually parsing the key and returning a map of string. Finally tests were added to both cover the new functionally as well the fixed bug.

```
[project_api_key]
projects/foo = your-api-key
^/home/user/projects/bar(\d+)/ = your-api-key
```

Closes #694


pkg/api/heartbeat.go
```diff
@@ -25,55 +26,139 @@ func (c *Client) SendHeartbeats(heartbeats []heartbeat.Heartbeat) ([]heartbeat.R
...
+	if len(grouped) == 1 {
+		c.sendHeartbeats(url, heartbeats, chres, cherr)
+
+		return <-chres, <-cherr
+	}
+
+	var wg sync.WaitGroup
+
+	for apiKey, hh := range grouped {
+		hh := hh
```
alanhamlett, line 47: Why do we re-define these?

alanhamlett (changes requested):
We need to run `apikey.WithReplacing` on offline heartbeats too.

pkg/api/heartbeat.go
gandarez: https://go.dev/doc/faq#closures_and_goroutines

pkg/vipertools/vipertools.go
```diff
@@ -60,3 +60,18 @@ func FirstNonEmptyString(v *viper.Viper, keys ...string) (string, bool) {
 func GetString(v *viper.Viper, key string) string {
 	return strings.Trim(v.GetString(key), `"'`)
 }
+
+// GetStringMapString gets a parameter/setting by key prefix and stripts any quotes.
+func GetStringMapString(v *viper.Viper, prefix string) map[string]string {
+	m := map[string]string{}
+
+	for _, k := range v.AllKeys() {
+		if !strings.HasPrefix(k, prefix) {
```
alanhamlett: This should be `if !strings.HasPrefix(k, prefix + ".") {` to prevent matching sections starting with prefix, for ex:

`[project_api_key_foo]`

pkg/api/heartbeat.go
alanhamlett: Ah, same as JavaScript. Thanks!

alanhamlett (approved):
Perfect!

The failing `Integration Tests Windows` is just flakey. After you squash it should pass.

PR #698, opened 2022-04-14
Downgrade checkout action

alanhamlett:
`actions/checkout@v3` clones the repo using the GitHub API, but semver-action needs it cloned with `git clone`.


gandarez:
Workflow already ran using `v3` with success.
https://github.com/wakatime/wakatime-cli/runs/5951396249?check_suite_focus=true

gandarez (approved):
let's give a try

PR #701, opened 2022-04-15
Parse ssh config file for sftp remote files

### LINKED ISSUE #658: Cannot read remote files through ssh
As discussed in https://github.com/wakatime/wakatime-cli/issues/426, waketime can not read remote files through ssh. I tried to make a minimal reproduction sample in this issue.

**Steps**
1. `ssh 124.16.139.208 -p 10722` login without password currectly.
2. `wakatime-cli --entity ssh://124.16.139.208:10722/home/wangtianshu/test_project/test.py` got error

**Configs**

`~/.ssh/config`: empty
`~/.ssh/knwon_hosts`:
```
[124.16.139.208]:10722 ssh-ed25519 AAAAC3***
[124.16.139.208]:10722 ssh-rsa AAAAB3***
[124.16.139.208]:10722 ecdsa-sha2-nistp256 AAAAE2***
```

**Logs**:
```
{"caller":"github.com/wakatime/wakatime-cli/pkg/remote/remote.go:88","func":"1","level":"error","message":"failed to download file to temporary folder: failed to connect to sftp host: failed to connecto to '124.16.139.208:10722': ssh: handshake failed: ssh: host key mismatch","now":"2022-02-23T10:35:58+08:00","version":"v1.37.0"}
{"caller":"github.com/wakatime/wakatime-cli/pkg/language/chroma.go:128","func":"selectByCustomizedPriority","level":"warning","message":"failed to load folder extensions: failed to read directory: open ssh://124.16.139.208:10722/home/wangtianshu/test_project/: no such file or directory","now":"2022-02-23T10:35:58+08:00","version":"v1.37.0"}
```

**Environment**:

- OS: mac
- Platform: 386

alanhamlett:
Fixes #658.

Parses `~/.ssh/config` for keys:

* `User` (user from url overrides config)
* `HostName`
* `Port` (value from url overrides config)
* `UserKnownHostsFile` for location of `~/.ssh/known_hosts`
* `StrictHostKeyChecking` and skips parsing known hosts file if set to `no`
* `IdentityFile` for location of `~/.ssh/id_rsa` file
* `HostKeyAlias` for looking up the known key


pkg/remote/remote.go
```diff
@@ -115,20 +145,40 @@ func NewClient(address string) (Client, error) {
...
+	if port == 0 {
+		port, _ = strconv.Atoi(ssh_config.Get(host, "Port"))
+	}
+
+	if port == 0 {
+		port, _ = strconv.Atoi(ssh_config.Get(derivedHost, "Port"))
+	}
+
+	if port == 0 {
+		port, _ = strconv.Atoi(ssh_config.Default("Port"))
```
gandarez: Don't think we need it here (_including str conversion_) since the default value is always port 22. So keeping the initial implementation (_with defaultPort_) seems to be better then this line.

pkg/remote/remote.go
```diff
@@ -115,20 +145,40 @@ func NewClient(address string) (Client, error) {
...
+
+	var port int
 
 	if parsedURL.Port() != "" {
 		// we're safe to ignore error here since `url.Parse` checks if port is valid integer
 		port, _ = strconv.Atoi(parsedURL.Port())
 		host = strings.Split(host, ":")[0]
 	}
 
+	derivedHost := ssh_config.Get(host, "HostName")
```
gandarez, line 142: What "derived host" mean and where does it come from? Any documentation out there for it?

pkg/remote/remote.go
```diff
@@ -25,19 +28,46 @@ import (
...
+	HostKeyAlias    string
+	HostDisplayName string
+	HostDerived     string
+	Port            int
+	Path            string
+}
+
+type HostSet map[string]struct{}
+
+func (i HostSet) Contains(hosts []string) bool {
```
gandarez: It's a simple iteration over an array of strings and don't need a method for it especially public. Method is when a struct is needed to call any func:
```Go
func (x struct) MyFunc(args..) bool {
    ....
}
```

I'd suggest:
```Go
func contains(values []string, args ...string) bool {
	for _, val := range values {
		for _, arg := range args {
			if val == arg {
				return true
			}
		}
	}

	return false
}
```

gandarez: I'd also move it to bottom since it has less importance then `WithDetection`, `NewClient`, etc...

pkg/remote/remote.go
```diff
@@ -25,19 +28,46 @@ import (
 var RemoteAddressRegex = regexp.MustCompile(`(?i)^((ssh|sftp)://)+(?P<credentials>[^:@]+(:([^:@])+)?@)?[^:]+(:\d+)?`)
 
 const (
-	defaultPort        = 22
 	defaultTimeoutSecs = 20
 	// Max file size supporting downloading from remote. Default is 512Kb.
 	maxFileSize = 512000
 )
 
+var defaultKeyFiles = []string{ // nolint:gochecknoglobals
```
gandarez:
```suggestion
// nolint:gochecknoglobals
var defaultKeyFiles = []string{
```

pkg/remote/remote.go
```diff
@@ -184,88 +407,104 @@ func (c Client) Connect() (*ssh.Client, *sftp.Client, error) {
...
+	alias := ssh_config.Get(host, "HostKeyAlias")
+	if alias == "" && host != hostDerived {
+		alias = ssh_config.Get(hostDerived, "HostKeyAlias")
+	}
+
+	if alias == "" {
+		return ""
+	}
+
+	alias, _ = homedir.Expand(alias)
```
gandarez: Not checking for error here.

pkg/remote/remote.go
```diff
@@ -166,13 +216,186 @@ func (c Client) DownloadFile(localFile string) error {
...
+
+	if c.HostDisplayName != c.HostDerived {
+		revokedKeysFile = ssh_config.Get(c.HostDerived, "RevokedHostKeys")
+		if revokedKeysFile != "" {
+			log.Warnln("Using ssh config RevokedHostKeys is not supported")
+		}
+	}
+}
+
+func (c Client) getSSHClient() (*ssh.Client, error) {
```
gandarez: Getter should not have `get` prefix https://go.dev/doc/effective_go#Getters.

pkg/remote/remote.go
```diff
@@ -184,88 +407,104 @@ func (c Client) Connect() (*ssh.Client, *sftp.Client, error) {
...
 
-			hostKey, _, _, _, err := ssh.ParseAuthorizedKey(scanner.Bytes())
-			if err != nil {
-				log.Warnf("failed to parse %q: %s", fields[2], err)
-				continue
-			}
+	return nil, fmt.Errorf("failed to connect to %s", addr)
+}
+
+func (c Client) getUser() string {
```
gandarez: Getter should not have get prefix https://go.dev/doc/effective_go#Getters.

pkg/remote/remote.go
```diff
@@ -184,88 +407,104 @@ func (c Client) Connect() (*ssh.Client, *sftp.Client, error) {
...
+		user := ssh_config.Get(c.HostDerived, "User")
+		if user != "" {
+			return user
+		}
+	}
+
+	return ""
+}
+
+func getHostKeyAlias(host string, hostDerived string) string {
```
gandarez: Getter should not have get prefix https://go.dev/doc/effective_go#Getters.

pkg/remote/remote.go
```diff
@@ -166,13 +216,186 @@ func (c Client) DownloadFile(localFile string) error {
...
+
+			break
+		}
+	}
+
+	if keyFile == "" {
+		return ""
+	}
+
+	keyFile, _ = homedir.Expand(keyFile)
```
gandarez: Not checking for error.

pkg/remote/remote.go
gandarez: Not checking for error.

pkg/remote/remote.go
```diff
@@ -166,13 +216,186 @@ func (c Client) DownloadFile(localFile string) error {
...
 
+	signer, err := ssh.ParsePrivateKey(key)
+	if err != nil {
+		log.Fatalf("unable to parse private key %s: %v", identityFile, err)
+	}
+
+	return signer
+}
+
+func (c Client) warnIfUsingRevokedHostKeys() {
```
gandarez, line 347: It seems we can do a small refactoring and remove this method. Let's discuss later.

pkg/remote/remote.go
```diff
@@ -166,13 +216,186 @@ func (c Client) DownloadFile(localFile string) error {
...
+	sftpClient, err := sftp.NewClient(sshClient)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to start sftp subsystem: %s", err)
+	}
+
+	return sshClient, sftpClient, nil
+}
+
+// getKnownHostKeys gets all host keys from local known hosts for given hosts.
+func (c Client) getKnownHostKeys() []ssh.PublicKey {
```
gandarez: Getter should not have get prefix https://go.dev/doc/effective_go#Getters.

pkg/remote/remote.go
```diff
@@ -166,13 +216,186 @@ func (c Client) DownloadFile(localFile string) error {
...
+
+	if strict == "accept-new" || strict == "off" {
+		strict = "no"
+	}
+
+	return strict
+}
+
+// getKnownHostsFile returns the path to the known hosts file.
+func (c Client) getKnownHostsFile() (string, error) {
```
gandarez: Getter should not have get prefix https://go.dev/doc/effective_go#Getters.

pkg/remote/remote.go
```diff
@@ -166,13 +216,186 @@ func (c Client) DownloadFile(localFile string) error {
...
+
+	if file == "" {
+		file = "~/.ssh/known_hosts"
+	}
+
+	return homedir.Expand(file)
+}
+
+// getIdentityFile returns the path to a secret key file, or the first existing default.
+func (c Client) getIdentityFile() string {
```
gandarez: Getter should not have get prefix https://go.dev/doc/effective_go#Getters.

gandarez (changes requested):

pkg/remote/remote.go
alanhamlett: It's the host defined in the `~/.ssh/config` file for the url's host. The host from the ssh config file overrides the host we connect to.

gandarez:
Also adding missing tests would be a good idea for the 400 lines changed and make sure everything works as expected.

PR #702, opened 2022-04-15
Allow importing another config file

### LINKED ISSUE #464: Include another configuration file
**Is your feature request related to a problem? Please describe.**

Normally, I have my configurations store as a repository on GitHub. And as you already know, storing passwords or secrets on public repository is a very bad practice. However, I would like to have my configurations stored on the repository still.

**Describe the solution you'd like to see**

Another configuration option (something like `config_file` or `extensions`, as `include` is already taken) for including another configuration file. So that I can keep my configurations separate from the secrets / API key. Or an option (something like `api_file`) to have it read / store from somewhere else is also fine for my use case.

**Describe alternatives you've considered**

Not yet explore, but so far this is the intuitive solution for me.

**Additional context**

None

gandarez:
This PR allows importing another config file other than the default `.wakatime.cfg` which may be used to store sensitive data outside home dir.

Closes #464


alanhamlett (approved):

alanhamlett:
A nice to have would be recursively parsing config files, so you could have a chain of multiple imported config files. However, it's not necessary in the first implementation.

PR #706, opened 2022-04-21
Use alternate project when no project detected

### LINKED ISSUE #270: WakaTime not automatically recognising VS Code workspaces as projects
As in the title -- it would be nice if they were, as it is more intentional to create a workspace than just to open a folder, and workspaces pretty much function as projects. If necessary, this could be an optional feature/setting, but it would save having to create a `.wakatime-project` for every workspace in its folder (with the same name as the workspace).

alanhamlett:
Fixes https://github.com/wakatime/vscode-wakatime/issues/270

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,8 +9,8 @@ To contribute to this project please carefully read this document.
 Prerequisites:
 
 - We use `make` to build and run tests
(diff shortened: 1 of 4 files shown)


SubjunctiveQuaver:
Thank you! Looking forward to the release :D

PR #708, opened 2022-04-21
Run integration tests synchronously

alanhamlett:
Fixes the flakey integration test `TestSendHeartbeats_ExtraHeartbeats` with `go test -p 1`.

(Renamed to `TestMain_SendHeartbeats_ExtraHeartbeats`)

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -49,11 +49,11 @@ func binaryPath(t *testing.T) string {
 	}
 }
 
-func TestSendHeartbeats(t *testing.T) {
+func TestMain_SendHeartbeats(t *testing.T) {
 	testSendHeartbeats(t, "testdata/main.go", "wakatime-cli")
 }
 
-func TestSendHeartbeats_EntityFileInTempDir(t *testing.T) {
+func TestMain_SendHeartbeats_EntityFileInTempDir(t *testing.T) {
 	tmpDir := t.TempDir()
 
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -49,11 +49,11 @@ func binaryPath(t *testing.T) string {
 	}
 }
 
-func TestSendHeartbeats(t *testing.T) {
+func TestMain_SendHeartbeats(t *testing.T) {
```
gandarez, line 52: The `Main` is implicitly defined in the package name `main_test`. The main purpose here is to describe which function is being tested and not the package.

gandarez (changes requested):
I'd nt rename tests as the `Main` is implicitly defined in the package name `main_test`. The main purpose here is to describe which function is being tested and not the package which is redundant.

gandarez:
Maybe this [PR](https://github.com/wakatime/wakatime-cli/pull/710) fixes it.

main_test.go
alanhamlett: I started renaming to make `TestSendHeartbeats_ExtraHeartbeats` uniquely searchable, since there's another test with that name. I'll revert it.

alanhamlett:
> Maybe this https://github.com/wakatime/wakatime-cli/pull/710 fixes it.

Yea, let's see. I'll close this and if it fails again in the future we can run sync.

PR #712, opened 2022-05-02
Log api key pattern match results in debug mode

alanhamlett:
To help with debugging why a `[project_api_key]` pattern is or isn't matching an entity path.


gandarez:
Is this somehow temporary?

alanhamlett:
No, I think we should make this permanent when debug is enabled.

alanhamlett:
Closing since this will be added when fixing #713.

PR #716, opened 2022-05-03
All user-input regex are case insensitive

### LINKED ISSUE #714: Pattern matching for include, exclude, and other regex patterns should be case insensitive
Any user-input regex pattern that matches file entity paths should be case-insensitive.

As part of this fix, let's remove [regex.go](https://github.com/wakatime/wakatime-cli/blob/cb7314a37fbd066e146a5b34149ef4369c8f7ea2/pkg/regex/regex.go) and use `regexp.Compile` instead.

gandarez:
This PR makes all user-input regex case insesitive by adding `(?i)` at the begining. Can't simply remove dependency to `github.com/dlclark/regexp2` due to lack of native support for negative/positive lookahead, affected testes are:

* TestLoadParams_Filter_Exclude_PerlRegexPatterns/negative_lookahead
* TestLoadParams_Filter_Exclude_PerlRegexPatterns/positive_lookahead
* TestLoadParams_Filter_Include_PerlRegexPatterns/negative_lookahead
* TestLoadParams_Filter_Include_PerlRegexPatterns/positive_lookahead

Closes https://github.com/wakatime/wakatime-cli/issues/714

### DIFF
--- a/cmd/params/params.go
+++ b/cmd/params/params.go
@@ -414,6 +414,11 @@ func loadFilterParams(v *viper.Viper) FilterParams {
 	var excludePatterns []regex.Regex
 
 	for _, s := range exclude {
+		// make all regex case insensitive
+		if !strings.HasPrefix(s, "(?i)") {
+			s = "(?i)" + s
+		}
+
 		compiled, err := regex.Compile(s)
 		if err != nil {
 			log.Warnf("failed to compile exclude regex pattern %q", s)
@@ -429,6 +434,11 @@ func loadFilterParams(v *viper.Viper) FilterParams {
 	var includePatterns []regex.Regex
 
 	for _, s := range include {
+		// make all regex case insensitive
+		if !strings.HasPrefix(s, "(?i)") {
+			s = "(?i)" + s
+		}
+
 		compiled, err := regex.Compile(s)
 		if err != nil {
 			log.Warnf("failed to compile include regex pattern %q", s)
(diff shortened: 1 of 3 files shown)


cmd/params/params.go
```diff
@@ -534,6 +544,11 @@ func loadProjectParams(v *viper.Viper) (ProjectParams, error) {
 	projectMap := vipertools.GetStringMapString(v, "projectmap")
 
 	for k, s := range projectMap {
+		// make all regex case insensitive
+		if !strings.HasPrefix(k, "(?i)") {
+			k = "(?i)" + k
+		}
+
 		compiled, err := regexp.Compile(k)
```
alanhamlett, line 552: Let's change this to:

```go
compiled, err := regex.Compile(k)
```

So we support negative lookahead for all user input regex patterns.

gandarez: Native lib doesn't support negative/positive lookahead. So if any user has already been using, this version will break for them.

alanhamlett: That's fine, the project map can always be fixed on the dashboard with custom rules. The ones that can't be fixed after sending are `include` and `exclude`.

alanhamlett: It's worth it to have all user-facing regex behave the same way.

alanhamlett (approved):

cmd/params/params.go
gandarez: I'll do it in a separate PR

PR #719, opened 2022-05-04
Use internal config when writing backoff

alanhamlett:
This fixes a bug where the `~/.wakatime.cfg` file was being used when writing `[internal]` backoff settings after an api error.

It also removes request header logging, because the API Key is logged fully as base64 text as part of the Authorization header.

Finally, it adds logging when api key pattern matching fails to match because we still have some bugs around it. We can remove that logging and only log when a match is found after the bugs are fixed.

### DIFF
--- a/pkg/api/heartbeat.go
+++ b/pkg/api/heartbeat.go
@@ -65,8 +65,6 @@ func (c *Client) sendHeartbeats(url string, heartbeats []heartbeat.Heartbeat) ([
 	}
 	defer resp.Body.Close()
 
-	log.Debugf("request headers: %+v", req.Header)
-
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, Err(fmt.Sprintf("failed reading response body from %q: %s", url, err))
--- a/pkg/apikey/apikey.go
+++ b/pkg/apikey/apikey.go
@@ -48,9 +48,11 @@ func WithReplacing(config Config) heartbeat.HandleOption {
 func MatchPattern(fp string, patterns []MapPattern) (string, bool) {
 	for _, pattern := range patterns {
 		if pattern.Regex.MatchString(fp) {
-			log.Debugf("api key matched %q", pattern.Regex.String())
+			log.Debugf("api key pattern %q matched path %q", pattern.Regex.String(), fp)
 			return pattern.ApiKey, true
 		}
+
+		log.Debugf("api key pattern %q did not match path %q", pattern.Regex.String(), fp)
 	}
 
 	return "", false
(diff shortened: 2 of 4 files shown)


gandarez:
Tests are probably failing due to the order of options passed for the handler in `cmd/offlinesync/offlinesync.go`. After you fix it I can review.

pkg/apikey/apikey.go
```diff
@@ -48,9 +48,10 @@ func WithReplacing(config Config) heartbeat.HandleOption {
 func MatchPattern(fp string, patterns []MapPattern) (string, bool) {
 	for _, pattern := range patterns {
 		if pattern.Regex.MatchString(fp) {
-			log.Debugf("api key matched %q", pattern.Regex.String())
+			log.Debugf("api key pattern '%q' matched path %q", pattern.Regex.String(), fp)
```
gandarez: `%q` already adds double quotes to the string.

pkg/apikey/apikey.go
```diff
@@ -48,9 +48,10 @@ func WithReplacing(config Config) heartbeat.HandleOption {
...
 			return pattern.ApiKey, true
 		}
+		log.Debugf("api key pattern '%q' did not match path %q", pattern.Regex.String(), fp)
```
gandarez: Same here for `'%q'`.

pkg/deps/deps.go
```diff
@@ -121,9 +121,6 @@ func Detect(filepath string, language heartbeat.Language) ([]string, error) {
 	case heartbeat.LanguageVBNet:
 		parser = &ParserVbNet{}
 	default:
-		log.Debugf(
```
gandarez, line 124: Isn't it worth having an "alert" for no parser found?

pkg/api/heartbeat.go
```diff
@@ -65,8 +65,6 @@ func (c *Client) sendHeartbeats(url string, heartbeats []heartbeat.Heartbeat) ([
 	}
 	defer resp.Body.Close()
 
-	log.Debugf("request headers: %+v", req.Header)
```
gandarez, line 68: oops! 👍

gandarez (changes requested):
left few remarks

pkg/api/heartbeat.go
alanhamlett: We can add logging for headers back if we loop over the headers and hide the `Authorization` header's value.

pkg/deps/deps.go
alanhamlett: Nope, I think it's fine to not have it. Maybe we'll add it back if we need it when adding more dependency language support in the future.

gandarez (approved):

PR #723, opened 2022-05-06
Print offline heartbeats to stdout

gandarez:
This PR adds new flag `--print-offline-heartbeats` to read bolt db and print heartbeats limited to maximum number informed.


cmd/params/params.go
```diff
@@ -599,9 +602,18 @@ func LoadOfflineParams(v *viper.Viper) (Offline, error) {
 		return Offline{}, errors.New("argument --sync-offline-activity must be \"none\" or a positive integer number")
 	}
 
+	switch {
+	case !v.IsSet("print-offline-heartbeats"):
```
alanhamlett: Do we need this? The `--print-offline-heartbeats` param requires an integer afterwards so it shouldn't ever need to use the default.

alanhamlett (approved):

cmd/params/params.go
gandarez: No we don't. I also have that fix in mind for `sync-offline-activity` as well.

gandarez: Fixed it and merged.

PR #724, opened 2022-05-08
Add missing tests for heartbeats count

gandarez:
Simply add missing tests for heartbeats count as well fix a minor bug where an error could happen and the returned value should be `zero`.

### DIFF
--- a/pkg/offline/offline.go
+++ b/pkg/offline/offline.go
@@ -339,20 +339,18 @@ func CountHeartbeats(filepath string) (int, error) {
 		return 0, fmt.Errorf("failed to start db transaction: %s", err)
 	}
 
+	defer func() {
+		err := tx.Rollback()
+		if err != nil {
+			log.Errorf("failed to rollback transaction: %s", err)
+		}
+	}()
+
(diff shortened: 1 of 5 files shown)


pkg/offline/offline.go
```diff
@@ -347,7 +347,7 @@ func CountHeartbeats(filepath string) (int, error) {
 
 		_ = tx.Rollback()
 
-		return count, err
+		return 0, err
```
alanhamlett: If there's an error then `Count()` would have already returned `0`.

alanhamlett (approved):

pkg/offline/offline.go
gandarez: In Go when error happens we would not trust/reuse the returned response. Think if the lib changes its behaviour to not return `0` but `-1` for example.

gandarez: I also improved code by defering rollback.
https://github.com/wakatime/wakatime-cli/compare/59ac5c08cb13a47e839988dded1fc15d21bb3083..00987f5fa86d8e8d911185c2210097d1aa03211d

PR #736, opened 2022-05-30
Ignore errors when parsing internal config file

alanhamlett:
Because the IDE plugins can write to `~/.wakatime-internal.cfg` at the same time as wakatime-cli, the file can become corrupted. This causes the plugin to stop tracking time and exit with a config parsing error. In this case, we want to just discard the internal configs and continue running.


alanhamlett:
No longer needed after #738.

PR #740, opened 2022-05-31
Skip unreconizable lines only for internal file

gandarez:
This PR sets skipping unreconizable lines **true** only for internal config file and allow the main one to fail if necessary. We also no longer need to call `fmt.Fprintf(os.Stderr, "....")` from `parseConfigFiles` since logging is being set before calling it.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	iniv1 "gopkg.in/ini.v1"
 )
 
 // Run executes commands parsed from a command line.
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -173,11 +188,18 @@ func parseConfigFiles(v *viper.Viper) error {
...
 		}
 
-		if err := ini.ReadInConfig(v, configFile); err != nil {
+		if err := ini.ReadInConfig(c.vp, configFile); err != nil {
 			fmt.Fprintf(os.Stderr, "failed to load configuration file: %s", err)
 
 			return fmt.Errorf("failed to load configuration file: %s", err)
 		}
+
+		if c.merge {
```
alanhamlett, line 193: Should we always merge?

Edit from offline convo: Merge happens later in `MergeConfig`. Just needed to manually merge because the new viper instance used to load internal config.

alanhamlett (approved):

PR #743, opened 2022-06-12
Backoff exit code 112 to prevent sending diagnostics when rate limited

alanhamlett:
We should skip sending diagnostics for expected errors, such as being rate limited.

### DIFF
--- a/cmd/offlinesync/offlinesync.go
+++ b/cmd/offlinesync/offlinesync.go
@@ -27,6 +27,7 @@ func Run(v *viper.Viper) (int, error) {
 	}
 
 	err = SyncOfflineActivity(v, queueFilepath)
+	// nolint:nestif
 	if err != nil {
 		var errauth api.ErrAuth
 		if errors.As(err, &errauth) {
@@ -40,6 +41,14 @@ func Run(v *viper.Viper) (int, error) {
 		if errors.As(err, &errbadRequest) {
 			return exitcode.ErrGeneric, fmt.Errorf(
 				"offline sync failed: bad request: %s",
+				err,
+			)
+		}
+
+		var errBackoff api.ErrBackoff
+		if errors.As(err, &errBackoff) {
+			return exitcode.ErrBackoff, fmt.Errorf(
+				"offline sync failed: rate limited: %s",
 				err,
 			)
 		}
--- a/main_test.go
+++ b/main_test.go
@@ -103,12 +103,18 @@ func testSendHeartbeats(t *testing.T, entity, project string) {
 
 	defer tmpFile.Close()
 
+	tmpInternalConfigFile, err := os.CreateTemp(tmpDir, "wakatime-internal.cfg")
+	require.NoError(t, err)
+
+	defer tmpInternalConfigFile.Close()
+
 	runWakatimeCli(
 		t,
 		&bytes.Buffer{},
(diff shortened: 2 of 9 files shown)


gandarez:
Pipeline is failing and maybe you need to change some lines to avoid nested ifs.

alanhamlett:
> Pipeline is failing and maybe you need to change some lines to avoid nested ifs.

Pipeline failed due to flakey integration tests, which seems to be fixed by always using temp files for `wakatime.cfg` and `wakatime-internal.cfg` to prevent tests accidentally affecting each other.

cmd/offlinesync/offlinesync.go
```diff
@@ -27,47 +27,55 @@ func Run(v *viper.Viper) (int, error) {
...
-			)
-		}
-
-		var errSyncDisabled ErrSyncDisabled
-		if errors.As(err, &errSyncDisabled) {
-			log.Debugln(err.Error())
-
-			return exitcode.Success, nil
-		}
+	if err == nil {
```
gandarez: In Go is recommended/good practice to first check if error happened and then do the rest of the code if no error. So in this case we need to simply add a new type assertion for the `ErrBackoff` inside the block `if err != nil`.

main_test.go
```diff
@@ -103,12 +103,18 @@ func testSendHeartbeats(t *testing.T, entity, project string) {
 
 	defer tmpFile.Close()
 
+	tmpInternalFile, err := os.CreateTemp(tmpDir, "wakatime-internal.cfg")
```
gandarez: What about renaming to `tmpInternalConfigFile` to have meaning?

main_test.go
```diff
@@ -375,15 +399,23 @@ func TestSendHeartbeats_MalformedConfig(t *testing.T) {
...
 func TestSendHeartbeats_MalformedInternalConfig(t *testing.T) {
-	offlineQueueFile, err := os.CreateTemp(t.TempDir(), "")
+	tmpDir := t.TempDir()
+
+	offlineQueueFile, err := os.CreateTemp(tmpDir, "")
 	require.NoError(t, err)
 
 	defer offlineQueueFile.Close()
 
+	tmpFile, err := os.CreateTemp(tmpDir, "wakatime.cfg")
```
gandarez, line 409: What about renaming to `tmpConfigFile` to have meaning?

gandarez (changes requested):
Left two comments regarding error handling and variable naming for `tmpInternalFile` and `tmpFile`.

cmd/offlinesync/offlinesync.go
alanhamlett: That caused linter to fail because the if condition was too complex.

alanhamlett: Ok, but I'll have to move the logic into a function because that way fails linter.

cmd/offlinesync/offlinesync.go
```diff
@@ -27,6 +27,7 @@ func Run(v *viper.Viper) (int, error) {
 	}
 
 	err = SyncOfflineActivity(v, queueFilepath)
+	// nolint:nestif
```
gandarez, line 30: 👍

gandarez (approved):

PR #744, opened 2022-06-12
Prevent ssh config tests from corrupting local ssh config file

alanhamlett:
Instead of overwriting local `~/.ssh/config` file with the testdata file, we mock out `ssh_config` to point it to a temp file.


go.mod
```diff
@@ -53,3 +53,5 @@ require (
 replace github.com/alecthomas/chroma => github.com/wakatime/chroma v0.10.0-wakatime.1
 
 replace github.com/matishsiao/goInfo => github.com/wakatime/goInfo v0.1.0-wakatime.8
+
+replace github.com/kevinburke/ssh_config => github.com/wakatime/ssh_config v1.2.0-wakatime.1
```
gandarez: Why do we need to replace this library?

pkg/remote/remote_test.go
```diff
@@ -106,22 +105,20 @@ func TestWithDetection_sshConfig_Hostname(t *testing.T) {
 	shutdown, host, port := testServer(t, false)
 	defer shutdown()
 
-	sshConfigFile, _ := homedir.Expand("~/.ssh/config")
+	tmpFile, err := os.CreateTemp(t.TempDir(), "")
+	require.NoError(t, err)
+
+	defer tmpFile.Close()
+
+	sshConfigFile := tmpFile.Name()
```
gandarez: It's being used once in the func so don't need to create a new variable for that, simply use `tmpFile.Name()` instead.

pkg/remote/remote_test.go
```diff
@@ -184,22 +181,20 @@ func TestWithDetection_sshConfig_UserKnownHostsFile_mismatch(t *testing.T) {
 	shutdown, host, port := testServer(t, true)
 	defer shutdown()
 
-	sshConfigFile, _ := homedir.Expand("~/.ssh/config")
+	tmpFile, err := os.CreateTemp(t.TempDir(), "")
+	require.NoError(t, err)
+
+	defer tmpFile.Close()
+
+	sshConfigFile := tmpFile.Name()
```
gandarez: same here

pkg/remote/remote_test.go
```diff
@@ -256,22 +251,20 @@ func TestWithDetection_sshConfig_UserKnownHostsFile_match(t *testing.T) {
...
```
gandarez: same here

gandarez (changes requested):
IMO we don't need to replace `ssh_config` if there isn't any bug fix or enhancement to the forked repo.

alanhamlett:
> IMO we don't need to replace `ssh_config` if there isn't any bug fix or enhancement to the forked repo.

We do need the latest change that makes `ConfigFinder` public so we can overwrite it with out mocked one.

go.mod
alanhamlett: We need this new public function recently added so we can mock the ssh config file location:
https://github.com/kevinburke/ssh_config/commit/5724bd127982289dd52e79933bbe60b98a5b639e#diff-0e426a43248661127a0c0ee115aef7a1093b635f8993b3f7ebb1dd9f05b8f249R267

alanhamlett: Updated to use master instead of fork.

gandarez (approved):
Looks good!

PR #747, opened 2022-06-18
Prevent downloading remote files when excluded

alanhamlett:
This PR prevents downloading remote files when the `entity` path matches an `exclude` filter, because we're throwing away the heartbeat anyway.

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -66,15 +66,6 @@ func Filter(h heartbeat.Heartbeat, config Config) error {
 		return fmt.Errorf(fmt.Sprintf("filter by pattern: %s", err))
 	}
 
-	if h.IsUnsavedEntity {
-		return nil
-	}
-
-	// filter file
-	if h.EntityType != heartbeat.FileType {
-		return nil
-	}
-
 	err := filterFileEntity(h, config)
 	if err != nil {
 		return fmt.Errorf(fmt.Sprintf("filter file: %s", err))
@@ -112,10 +103,22 @@ func filterByPattern(entity string, include, exclude []regex.Regex) error {
 // the existence of the passed in filepath, and optionally by checking if a
 // wakatime project file can be detected in the filepath directory tree.
 // Returns an error to signal to the caller to skip the heartbeat.
-func filterFileEntity(heartbeat heartbeat.Heartbeat, config Config) error {
-	entity := heartbeat.Entity
-	if heartbeat.LocalFile != "" {
-		entity = heartbeat.LocalFile
+func filterFileEntity(h heartbeat.Heartbeat, config Config) error {
+	if h.EntityType != heartbeat.FileType {
+		return nil
+	}
+
+	if h.IsUnsavedEntity {
+		return nil
+	}
+
+	if h.IsRemote() {
+		return nil
+	}
+
+	entity := h.Entity
+	if h.LocalFile != "" {
+		entity = h.LocalFile
 	}
 
 	// skip files that don't exist on disk
@@ -123,9 +126,8 @@ func filterFileEntity(heartbeat heartbeat.Heartbeat, config Config) error {
 		return fmt.Errorf(fmt.Sprintf("skipping because of non-existing file %q", entity))
 	}
 
-	// when including only with project file, skip files when the project does not contain a .wakatime-project file
-	// except when file was downloaded to a tmp local file
-	if config.IncludeOnlyWithProjectFile && heartbeat.LocalFile == "" {
+	// when including only with project file, skip files when the project doesn't have a .wakatime-project file
+	if config.IncludeOnlyWithProjectFile {
 		_, ok := project.FindFileOrDirectory(entity, project.WakaTimeProjectFile)
 		if !ok {
 			return fmt.Errorf("skipping because missing .wakatime-project file in parent path")
--- a/pkg/filter/filter_test.go
+++ b/pkg/filter/filter_test.go
@@ -175,13 +175,13 @@ func TestFilter_ExistingProjectFile(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestFilter_RemoteFile(t *testing.T) {
+func TestFilter_RemoteFileSkipsFiltering(t *testing.T) {
 	h := testHeartbeat()
 	h.LocalFile = h.Entity
 	h.Entity = "ssh://wakatime:1234@192.168.1.1/path/to/remote/main.go"
 
 	err := filter.Filter(h, filter.Config{})
-	assert.EqualError(t, err, "filter file: skipping because of non-existing file \"/tmp/main.go\"")
+	require.NoError(t, err)
 }
 
 func TestFilter_ErrNonExistingProjectFile(t *testing.T) {
--- a/pkg/heartbeat/heartbeat.go
+++ b/pkg/heartbeat/heartbeat.go
@@ -3,6 +3,7 @@ package heartbeat
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -12,6 +13,9 @@ import (
 
 	"github.com/matishsiao/goInfo"
 )
+
+// remoteAddressRegex is a pattern for (ssh|sftp)://user:pass@host:port.
+var remoteAddressRegex = regexp.MustCompile(`(?i)^((ssh|sftp)://)+(?P<credentials>[^:@]+(:([^:@])+)?@)?[^:]+(:\d+)?`)
 
 // Heartbeat is a structure representing activity for a user on a some entity.
 type Heartbeat struct {
@@ -108,6 +112,19 @@ func (h Heartbeat) ID() string {
 		h.Entity,
 		isWrite,
 	)
+}
+
+// IsRemote returns true when entity is a remote file.
+func (h Heartbeat) IsRemote() bool {
+	if h.EntityType != FileType {
+		return false
+	}
+
+	if h.IsUnsavedEntity {
+		return false
+	}
+
+	return remoteAddressRegex.MatchString(h.Entity)
 }
 
 // Result represents a response from the wakatime api.
--- a/pkg/heartbeat/heartbeat_test.go
+++ b/pkg/heartbeat/heartbeat_test.go
@@ -243,6 +243,42 @@ func TestPluginFromUserAgent(t *testing.T) {
 	}
 }
 
+func TestRemoteAddressRegex(t *testing.T) {
+	tests := map[string]struct {
+		Heartbeat heartbeat.Heartbeat
+		Expected  bool
+	}{
+		"ssh full path": {
+			Heartbeat: heartbeat.Heartbeat{Entity: "ssh://user:1234@192.168.1.2/home/pi/unicorn-hat/examples/ascii_pic.py"},
+			Expected:  true,
+		},
+		"sftp full path": {
+			Heartbeat: heartbeat.Heartbeat{Entity: "sftp://user:1234@192.168.1.2/home/pi/unicorn-hat/examples/ascii_pic.py"},
+			Expected:  true,
+		},
+		"without path": {
+			Heartbeat: heartbeat.Heartbeat{Entity: "ssh://user:1234@192.168.1.2"},
+			Expected:  true,
+		},
+		"invalid ftp": {
+			Heartbeat: heartbeat.Heartbeat{Entity: "ftp://user:1234@192.168.1.2"},
+			Expected:  false,
+		},
+		"invalid": {
+			Heartbeat: heartbeat.Heartbeat{Entity: "http://192.168.1.2"},
+			Expected:  false,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			result := test.Heartbeat.IsRemote()
+
+			assert.Equal(t, test.Expected, result)
+		})
+	}
+}
+
 type mockSender struct {
 	SendHeartbeatsFn        func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error)
 	SendHeartbeatsFnInvoked bool
--- a/pkg/offline/offline.go
+++ b/pkg/offline/offline.go
@@ -67,7 +67,7 @@ func WithQueue(filepath string) heartbeat.HandleOption {
 				requeueErr := pushHeartbeatsWithRetry(filepath, hh)
 				if requeueErr != nil {
 					return nil, fmt.Errorf(
-						"failed to push heatbeats to queue after api error: %s. error: %s",
+						"failed to push heartbeats to queue after api error: %s. error: %s",
 						requeueErr,
 						err)
 				}
--- a/pkg/project/filter.go
+++ b/pkg/project/filter.go
@@ -2,6 +2,7 @@ package project
 
 import (
 	"fmt"
+	"os"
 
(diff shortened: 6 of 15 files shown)


pkg/heartbeat/heartbeat.go
```diff
@@ -109,6 +111,22 @@ func (h Heartbeat) ID() string {
 	)
 }
 
+// RemoteAddressRegex is a pattern for (ssh|sftp)://user:pass@host:port.
+var RemoteAddressRegex = regexp.MustCompile(`(?i)^((ssh|sftp)://)+(?P<credentials>[^:@]+(:([^:@])+)?@)?[^:]+(:\d+)?`)
```
gandarez: Let's move var declaration to top right after "import" clause.

pkg/filter/filter.go
```diff
@@ -103,6 +103,11 @@ func filterByPattern(entity string, include, exclude []regex.Regex) error {
 // wakatime project file can be detected in the filepath directory tree.
 // Returns an error to signal to the caller to skip the heartbeat.
 func filterFileEntity(heartbeat heartbeat.Heartbeat, config Config) error {
+	// skip filters for remote files downloaded to a local tmp file
+	if heartbeat.IsRemote() {
```
gandarez: It could be moved upwards to replace lines 59-66 in this file.
```suggestion
func Filter(h heartbeat.Heartbeat, config Config) error {
    // omitted for brevity
    
    // skip filters for remote files downloaded to a local tmp file
    if heartbeat.IsRemote() {
	return nil
    }
    
    err := filterFileEntity(h, config)
    // omitted for brevity
}
```

pkg/heartbeat/heartbeat_test.go
```diff
@@ -243,6 +243,42 @@ func TestPluginFromUserAgent(t *testing.T) {
 	}
 }
 
+func TestRegex(t *testing.T) {
```
gandarez: Let's rename to `TestRemoteAddressRegex`

pkg/heartbeat/heartbeat.go
```diff
@@ -30,6 +31,7 @@ type Heartbeat struct {
 	LineNumber          *int       `json:"lineno"`
 	Lines               *int       `json:"lines"`
 	LocalFile           string     `json:"-"`
+	LocalFileTemporary  bool       `json:"-"`
```
gandarez: That can be safely removed since it's direct related to `LocalFile != ""`, isn't it? I wouldn't set a control variable for another variable unless it adds a behaviour to `LocalFile` such as (_but I also don't like this approach_):
```Go
func (h Heartbeat) LocalFileTemporary() bool {
    return h.LocalFile != ""
}
```

I'd go simple by checking for `LocalFile` is not empty.

pkg/remote/remote.go
```diff
@@ -91,19 +81,48 @@ func WithDetection() heartbeat.HandleOption {
...
+// WithCleanup initializes and returns a heartbeat handle option, which
+// deletes a local temporary file if downloaded from a remote file.
+func WithCleanup() heartbeat.HandleOption {
+	return func(next heartbeat.Handle) heartbeat.Handle {
+		return func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
+			log.Debugln("execute remote cleanup")
+
+			for _, h := range hh {
+				if h.LocalFileTemporary {
+					os.Remove(h.LocalFile)
```
gandarez: We might check if error happened during clean up and log as Debug or Warn.

pkg/project/filter.go
```diff
@@ -28,6 +29,10 @@ func WithFiltering(config FilterConfig) heartbeat.HandleOption {
 				if err != nil {
 					log.Debugln(err.Error())
 
+					if h.LocalFileTemporary {
```
gandarez: It will never be true because `remote.WithDetection()` always run after `filter.WithFiltering()`.

pkg/remote/remote_test.go
```diff
@@ -295,16 +263,90 @@ func TestWithDetection_sshConfig_UserKnownHostsFile_match(t *testing.T) {
...
+			EntityType: heartbeat.FileType,
+			Time:       1585598060,
+			UserAgent:  "wakatime/13.0.7",
+		},
+	})
+	require.NoError(t, err)
+	assert.Len(t, results, 1)
+}
+
+func TestWithDetection_filtered(t *testing.T) {
```
gandarez:
```suggestion
func TestWithDetection_Filtered(t *testing.T) {
```

pkg/remote/remote_test.go
```diff
@@ -314,6 +356,81 @@ func TestWithDetection_sshConfig_UserKnownHostsFile_match(t *testing.T) {
...
+			LocalFile:          tmpFile.Name(),
+			LocalFileTemporary: true,
+		},
+	})
+	require.NoError(t, err)
+
+	assert.False(t, fileExists(tmpFile.Name()))
+}
+
+func TestWithCleanup_not_temporary(t *testing.T) {
```
gandarez:
```suggestion
func TestWithCleanup_NotTemporary(t *testing.T) {
```

gandarez (changes requested):
I'm tending to believe we don't need to check twice (_or even more times_) if the quantity of heartbeats is equal to zero to abort handling. Instead I'd create a new handle option to check the size and then abort.

pkg/heartbeat/heartbeat.go
alanhamlett: No, sometimes local file is passed from `--local-file` but it might not be a temporary file so we shouldn't delete it.

pkg/project/filter.go
alanhamlett: This is `project.WithFiltering` not `filter.WithFiltering`.

pkg/heartbeat/heartbeat.go
```diff
@@ -12,6 +13,9 @@ import (
 	"github.com/matishsiao/goInfo"
 )
 
+// RemoteAddressRegex is a pattern for (ssh|sftp)://user:pass@host:port.
+var RemoteAddressRegex = regexp.MustCompile(`(?i)^((ssh|sftp)://)+(?P<credentials>[^:@]+(:([^:@])+)?@)?[^:]+(:\d+)?`)
```
gandarez: It doesn't need to be exported. Let's keep it unexported since `IsRemote()` privately calls it and is already exported. Or it could be kept in `remote` package as is.

pkg/filter/filter.go
```diff
@@ -107,30 +123,3 @@ func filterByPattern(entity string, include, exclude []regex.Regex) error {
 
 	return nil
 }
-
-// filterFileEntity determines if a heartbeat of type file should be skipped, by verifying
-// the existence of the passed in filepath, and optionally by checking if a
-// wakatime project file can be detected in the filepath directory tree.
-// Returns an error to signal to the caller to skip the heartbeat.
-func filterFileEntity(heartbeat heartbeat.Heartbeat, config Config) error {
```
gandarez, line 115: I wouldn't remove this func beause it has a meaning and is well described. Maybe check if is remote in it and describe that in the func comments?

pkg/filter/filter_test.go
```diff
@@ -66,7 +65,7 @@ func TestWithFiltering(t *testing.T) {
 func TestWithLengthValidator(t *testing.T) {
 	opt := filter.WithLengthValidator()
 	h := opt(func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
-		return []heartbeat.Result{}, errors.New("this will should never be called")
+		return []heartbeat.Result{}, nil
```
gandarez: Why `error` was removed?

pkg/remote/remote.go
```diff
@@ -134,6 +127,11 @@ func WithCleanup() heartbeat.HandleOption {
 	}
 }
 
+func removeWithLogging(file string) {
```
gandarez: Let's rename it to `deleteLocalFile`? It has a bug where error is not being validated and always being logged.

pkg/heartbeat/heartbeat.go
```diff
@@ -30,6 +34,7 @@ type Heartbeat struct {
 	LineNumber          *int       `json:"lineno"`
 	Lines               *int       `json:"lines"`
 	LocalFile           string     `json:"-"`
+	LocalFileTemporary  bool       `json:"-"`
```
gandarez: Thre's already a dedicated field called `EntityRaw` which is only populated when a remote file is downloaded. Simply checking if `EntityRaw is not empty` does the job.

pkg/project/filter.go
gandarez: It already calls `remote.WithCleanup()` at the end of the chain, so is it needed?

pkg/remote/remote.go
```diff
@@ -91,22 +81,28 @@ func WithDetection() heartbeat.HandleOption {
...
+					removeWithLogging(tmpFile.Name())
+
 					continue
 				}
 
-				hh[i].LocalFile = tmpFile.Name()
-				// we save untouched entity for offline handling
-				hh[i].EntityRaw = h.Entity
+				h.LocalFile = tmpFile.Name()
+				h.LocalFileTemporary = true
```
gandarez: Not needed?!

pkg/remote/remote_test.go
```diff
@@ -211,12 +179,11 @@ func TestWithDetection_SshConfig_UserKnownHostsFile_Mismatch(t *testing.T) {
...
 	opts := []heartbeat.HandleOption{
-		remote.WithDetection(),
 		filter.WithFiltering(filter.Config{
-			Exclude:                    nil,
-			Include:                    nil,
-			IncludeOnlyWithProjectFile: false,
+			IncludeOnlyWithProjectFile: true,
 		}),
+		remote.WithDetection(),
+		remote.WithCleanup(),
```
gandarez: The `testing` package takes care of deleting temp files and directories at the end. So we can safely remove `remote.WithCleanup()`.

pkg/remote/remote_test.go
```diff
@@ -295,16 +263,17 @@ func TestWithDetection_SshConfig_UserKnownHostsFile_Match(t *testing.T) {
...
 
 	opts := []heartbeat.HandleOption{
-		remote.WithDetection(),
 		filter.WithFiltering(filter.Config{
 			Exclude:                    nil,
 			Include:                    nil,
 			IncludeOnlyWithProjectFile: true,
 		}),
+		remote.WithDetection(),
+		remote.WithCleanup(),
```
gandarez: same here.

gandarez (changes requested):

pkg/filter/filter_test.go
alanhamlett: Oops, I put it back now.

pkg/project/filter.go
alanhamlett: Yes, because the heartbeat is filtered out from the array and won't exist when `remote.WithCleanup` runs.

Instead, we could add a field to heartbeat: `Skip` or `IsFiltered`. Then set that boolean true instead of removing heartbeats from the array?

gandarez: That's make more sense to me.

alanhamlett: I added `IsFiltered` and it's working well.

alanhamlett: I had to revert back to removing filtered heartbeats from the array, because it causes unintended side-effects like downloading filtered remote files, calculating file stats on filtered files, detecting language of filtered files, etc.

Removing the heartbeat is better than checking `heartbeat.IsFiltered` in every heartbeat handler.

pkg/offline/offline.go
```diff
@@ -67,7 +67,7 @@ func WithQueue(filepath string) heartbeat.HandleOption {
 				requeueErr := pushHeartbeatsWithRetry(filepath, hh)
 				if requeueErr != nil {
 					return nil, fmt.Errorf(
-						"failed to push heatbeats to queue after api error: %s. error: %s",
+						"failed to push heartbeats to queue after api error: %s. error: %s",
```
gandarez, line 70: 👍

pkg/remote/remote.go
```diff
@@ -91,22 +81,27 @@ func WithDetection() heartbeat.HandleOption {
...
+					deleteLocalFile(tmpFile.Name())
+
 					continue
 				}
 
-				hh[i].LocalFile = tmpFile.Name()
-				// we save untouched entity for offline handling
-				hh[i].EntityRaw = h.Entity
+				h.LocalFile = tmpFile.Name()
+				h.EntityRaw = h.Entity // save untouched entity for offline handling
```
gandarez, line 99: In Go the coments are usually put before the declaration or the block code as we had before this line changes.

pkg/remote/remote_test.go
```diff
@@ -211,12 +177,11 @@ func TestWithDetection_SshConfig_UserKnownHostsFile_Mismatch(t *testing.T) {
...
 	opts := []heartbeat.HandleOption{
-		remote.WithDetection(),
 		filter.WithFiltering(filter.Config{
-			Exclude:                    nil,
-			Include:                    nil,
-			IncludeOnlyWithProjectFile: false,
+			IncludeOnlyWithProjectFile: true,
 		}),
+		remote.WithDetection(),
+		filter.WithLengthValidator(),
```
gandarez: Is that really needed for the test to work? Doesn't that piece of code produce the expected result?
```Go
sender := mockSender{
	SendHeartbeatsFn: func(hh []heartbeat.Heartbeat) ([]heartbeat.Result, error) {
		assert.Empty(t, hh)
		return []heartbeat.Result{}, nil
	},
}
```

pkg/remote/remote_test.go
```diff
@@ -295,16 +260,84 @@ func TestWithDetection_SshConfig_UserKnownHostsFile_Match(t *testing.T) {
...
+	}
+
+	opts := []heartbeat.HandleOption{
+		filter.WithFiltering(filter.Config{
+			Exclude:                    []regex.Regex{regexp.MustCompile(".*")},
+			Include:                    nil,
+			IncludeOnlyWithProjectFile: true,
+		}),
+		remote.WithDetection(),
+		filter.WithLengthValidator(),
```
gandarez: Same here.

pkg/remote/remote_test.go
```diff
@@ -314,6 +347,44 @@ func TestWithDetection_SshConfig_UserKnownHostsFile_Match(t *testing.T) {
...
+					Status:    201,
+					Heartbeat: heartbeat.Heartbeat{},
+				},
+			}, nil
+		},
+	}
+
+	handle := heartbeat.NewHandle(&sender, opts...)
+
+	assert.True(t, fileExists(tmpFile.Name()))
```
gandarez: Let's use `assert.FileExists()`.

pkg/remote/remote_test.go
```diff
@@ -314,6 +347,44 @@ func TestWithDetection_SshConfig_UserKnownHostsFile_Match(t *testing.T) {
...
+
+	_, err = handle([]heartbeat.Heartbeat{
+		{
+			LocalFile: tmpFile.Name(),
+		},
+	})
+	require.NoError(t, err)
+
+	assert.True(t, fileExists(tmpFile.Name()))
```
gandarez: Same here.

gandarez (changes requested):

gandarez (approved):
💪

PR #748, opened 2022-06-22
Abort heartbeat handling if list is empty

gandarez:
This PR adds a new handle option `filter.WithLengthValidator()` to abort heartbeat handling if list is empty. It has been added at the end of the chain to make sure all other filters/modifiers have been run.

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -243,6 +243,7 @@ func initHandleOptions(params paramscmd.Params) []heartbeat.HandleOption {
 			ProjectPatterns:      params.Heartbeat.Sanitize.HideProjectNames,
 			RemoteAddressPattern: remote.RemoteAddressRegex,
 		}),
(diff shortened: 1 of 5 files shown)


alanhamlett (approved):

alanhamlett:
The `TestWithFiltering_AbortAllFiltered` test needs to return `nil` like:
https://github.com/wakatime/wakatime-cli/pull/747/files#diff-ea7e66a8e0ac2e98df1773bc059d61fd0377493fa7d0f3da27dbdcacc8e8fd65R68

gandarez:
> The `TestWithFiltering_AbortAllFiltered` test needs to return `nil` like: https://github.com/wakatime/wakatime-cli/pull/747/files#diff-ea7e66a8e0ac2e98df1773bc059d61fd0377493fa7d0f3da27dbdcacc8e8fd65R68

That one was replaced by `TestWithLengthValidator` since `WithDiltering` won't abort heartbeat handling anymore.

PR #749, opened 2022-06-22
Cleanup temporary directories for remote files

gandarez:
This PR adds `remote.WithCleanup()` to make sure any file that was downloaded during remote detection has been purged. It uses `EntityRaw` attribute from Heartbeat since it's only set when a remote file is downloaded.


alanhamlett:
Let's rename `EntityRaw` to `EntityRemote`?

alanhamlett (approved):

PR #750, opened 2022-06-22
Remove unnecessary else

hezhizhen:


gandarez:
@hezhizhen thanks for opening this PR. Please read our `CONTRIBUTING` document [here](https://github.com/wakatime/wakatime-cli/blob/develop/CONTRIBUTING.md#branching-stratgegy) and [here](https://github.com/wakatime/wakatime-cli/blob/develop/CONTRIBUTING.md#pull-requests) to get this PR approved. It also has failing tests because linting problems.

pkg/ini/ini.go
```diff
@@ -183,11 +183,10 @@ func WakaHomeDir() (string, error) {
 	home, exists := os.LookupEnv("WAKATIME_HOME")
 	if exists && home != "" {
 		home, err := homedir.Expand(home)
-		if err != nil {
```
alanhamlett, line 186: I think the convention is to always check `if err != nil` and never `if err == nil`?

hezhizhen:
> @hezhizhen thanks for opening this PR. Please read our `CONTRIBUTING` document [here](https://github.com/wakatime/wakatime-cli/blob/develop/CONTRIBUTING.md#branching-stratgegy) and [here](https://github.com/wakatime/wakatime-cli/blob/develop/CONTRIBUTING.md#pull-requests) to get this PR approved. It also has failing tests because linting problems.

Sorry for replying late. I'll fix it later.

pkg/ini/ini.go
hezhizhen: Not always. The key is to return short cases first.

alanhamlett:
Implemented with #755.

PR #751, opened 2022-07-07
Detect Android operating system when running in Termux

alanhamlett:
When running in [Termux](https://play.google.com/store/apps/details?id=com.termux&hl=en_US&gl=US), for example when coding with an Android Tablet, the output from `wakatime-cli --user-agent` should look like:

`wakatime/v1.49.0 (Android-linux-5.4.0-104-generic-x86_64) go1.18.3 vim/800 vim-wakatime/9.0.1`

Previously, it would not include `Android` in the user agent string:

`wakatime/v1.49.0 (linux-5.4.0-104-generic-x86_64) go1.18.3 vim/800 vim-wakatime/9.0.1`

### DIFF
--- a/pkg/heartbeat/heartbeat.go
+++ b/pkg/heartbeat/heartbeat.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/wakatime/wakatime-cli/pkg/log"
+	"github.com/wakatime/wakatime-cli/pkg/system"
 	"github.com/wakatime/wakatime-cli/pkg/version"
 
 	"github.com/matishsiao/goInfo"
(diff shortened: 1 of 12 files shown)


pkg/heartbeat/osname_linux.go
```diff
@@ -0,0 +1,33 @@
...
+
+package heartbeat
+
+import (
+	"syscall"
+
+	"github.com/wakatime/wakatime-cli/pkg/log"
+)
+
+func osName() string {
```
gandarez: This function is not related to heartbeat so let's move to another package. I think a good descriptor could be `system` (any suggestion?). Then it would look like `system.OsName()`.

pkg/heartbeat/heartbeat.go
```diff
@@ -162,10 +156,21 @@ func UserAgent(plugin string) string {
 		log.Debugf("goInfo.GetInfo error: %s", err)
 	}
 
+	os := runtime.GOOS
+	alternateOS := osName()
```
gandarez: If `osName()` is moved to a more meanful package then this function can embed the logic for `runtime.GOOS` and there's no need to do `if alternateOS != "" && !strings.EqualFold(alternateOS, os)` here.

gandarez (changes requested):
I believe `osName()` does not fit well `heartbeat` package so I suggested few changes to better accomodate it.

gandarez (approved):
Looks good now, thanks.

PR #754, opened 2022-07-11
We dont need EntityRaw

alanhamlett:
When pushing heartbeats to the offline db, we should save the entity after filtering & formatting. The same entity that we send to the API should be saved to the offline db, not the original entity from `EntityRaw`. That's because when we send offline heartbeats to API, we don't re-run filtering & formatting. If we save `EntityRaw` to offline db, then we miss filtering & formatting on that entity when sending to API later.

Also implements #750.


alanhamlett:
Closing in favor of #755, because it's a bugfix not misc change.

PR #760, opened 2022-08-01
Add cloudfare resolver as a fallback

### LINKED ISSUE #664: DNS lookup failure on Arch Linux
- it should update the status of my coding... but as you can see in below pictures, it cannot do that.
- Api error / cannot update anything to the server:

image

**Environment**:

- OS: Arch Linux
- Platform: amd64 / x86_64

**Logs**:
image

Any suggestions?

gandarez:
This PR adds `cloudfare` as a fallback name resolver to fix DNS lookup when `connection refused` is returned.

Fix #664

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1,7 +1,11 @@
 package api
 
 import (
+	"errors"
+	"net"
 	"net/http"
+
+	"github.com/wakatime/wakatime-cli/pkg/log"
 )
 
 const (
@@ -35,8 +39,7 @@ func NewClient(baseURL string, opts ...Option) *Client {
 	c := &Client{
 		baseURL: baseURL,
 		client: &http.Client{
-			Transport:     NewTransport(),
-			CheckRedirect: nil, // defaults to following up to 10 redirects
+			Transport: NewTransport(),
 		},
 		doFunc: func(c *Client, req *http.Request) (*http.Response, error) {
 			req.Header.Set("Accept", "application/json")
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -54,5 +57,20 @@ func NewClient(baseURL string, opts ...Option) *Client {
 // Do executes c.doFunc(), which in turn allows wrapping c.client.Do() and manipulating
 // the request behavior of the api client.
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
-	return c.doFunc(c, req)
+	resp, err := c.doFunc(c, req)
+	if err != nil {
+		if errors.Is(err, syscall.ECONNREFUSED) {
```
alanhamlett: Will this match connection errors from `dns lookup` as well as errors connecting to `api.wakatime.com` servers?

We want to prevent using alternate DNS resolver when the connection fails to `api.wakatime.com`, and only use alternate DNS resolver when the `DNS lookup` fails.

pkg/api/transport.go
```diff
@@ -11,15 +13,36 @@ import (
...
+	t := NewTransport()
+
+	t.DialTLSContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		r := &net.Resolver{
+			PreferGo: false,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				d := net.Dialer{
+					Timeout: DefaultTimeoutSecs * time.Second,
+				}
+				return d.DialContext(ctx, network, "1.1.1.1:53")
```
alanhamlett, line 43: We can't have multiple resolvers right? Some people are on ipv6-only networks and the ipv4 address will never work for them.

pkg/api/api.go
alanhamlett: Let's also check for `dial tcp: lookup` in the error message string before using alternate DNS resolver?

https://stackoverflow.com/a/67647035/1290627

pkg/api/api.go
```diff
@@ -54,5 +57,20 @@ func NewClient(baseURL string, opts ...Option) *Client {
...
+			log.Warnf("connection refused: %s. It will retry with cloudfare dns", req.URL)
```
alanhamlett: Change to:

`log.Warnf("dns lookup failed: %s. Retrying with fallback dns resolver", req.URL)`

alanhamlett (changes requested):

gandarez:
@alanhamlett I changed to check for `*net.DNSError` and also added ipv6 support.

alanhamlett (approved):
Perfect!

PR #766, opened 2022-08-22
Bump Golang to 1.19

gandarez:


alanhamlett:
What do we gain and what do we lose from this upgrade? From Go 1.8 release notes, it looks like we [lose OS X 10.8 support](https://go.dev/doc/go1.8#ports) but I don't see that mentioned in [Go 1.9 release notes](https://go.dev/doc/go1.9#ports).

gandarez:
@alanhamlett here's a summary what has been improved in v1.19

https://www.sobyte.net/post/2022-06/go-1-19/

alanhamlett:
Main thing in worried about is dropping support for older operating systems, but looks like nothing changed there in this release.

alanhamlett (approved):

PR #771, opened 2022-09-08
Use host ip address when dns fails

gandarez:
This PR replaces the cloudfare DNS by WakaTime API ip address when DNS fails to resolve it.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -2,8 +2,10 @@ package api
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
+	"strings"
 
 	"github.com/wakatime/wakatime-cli/pkg/log"
 )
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -75,3 +83,17 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 
 	return resp, nil
 }
+
+func isLocalIPv6() bool {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
```
alanhamlett: Let's use `baseIPAddrv4` instead of `8.8.8.8` since the actual IP doesn't matter.

alanhamlett:
One last thing, do we need to prevent using the baseIPAddr when a custom `api_url` is defined by the user?

gandarez:
> One last thing, do we need to prevent using the baseIPAddr when a custom `api_url` is defined by the user?

Yes it should avoid overwriting host address.

alanhamlett:
Getting an SSL cert error when testing locally. Looks like we need to disable the Host verification part of cert verification only when overwriting the host with IP:

```
{"caller":"pkg/api/api.go:83","func":"api.(*Client)","level":"warning","message":"dns error, it will retry with host ip address '143.244.210.202'","now":"2022-09-08T23:45:19+02:00","version":"<local-build>"}
{"caller":"cmd/run.go:270","func":"cmd.runCmd","level":"error","message":"failed to run command: today fetch failed: failed fetching today from api: failed to make request to \"https://api.wakatime.com/api/v1/users/current/statusbar/today\": Get \"https://143.244.210.202/api/v1/users/current/statusbar/today\": x509: cannot validate certificate for 143.244.210.202 because it doesn't contain any IP SANs","now":"2022-09-08T23:45:30+02:00","version":"<local-build>"}
```

alanhamlett (approved):

PR #788, opened 2022-11-12
Ensure saving offline heartbeats finish before main thread exit

gandarez:
This PR adds a channel to the goroutine to make sure it finishes before main thread exit.


alanhamlett (approved):
Can we unit test this?

gandarez:
It's being tested with `TestSendHeartbeats_ExtraHeartbeats` in integration tests.

alanhamlett:
It's not fully tested because `TestSendHeartbeats_ExtraHeartbeats` was passing before this change. It's best when fixing bugs like this to have a new test that fails without the fix, and passes with the fix.

PR #797, opened 2022-12-29
More lax extra heartbeats parsing

alanhamlett:
Was seeing a lot of these error messages, meaning the JSON was probably valid but being discarded so we lost extra heartbeats:

```
{"caller":"cmd/params/params.go:395","func":"params.LoadHeartbeatParams","level":"error","message":"failed to read extra heartbeats: failed to read data from stdin: EOF","now":"2022-12-03T21:47:03+01:00","os/arch":"darwin/amd64","version":"v1.60.1"}
```

We should continue trying to parse the JSON even when `in.ReadString('\n')` returns an error because it still could return valid JSON data.


gandarez (approved):
👍

PR #815, opened 2023-01-23
Omit heartbeat nil fields

gandarez:
This PR removes all nullables fields from the heartbeat payload to reduce its size and reduce network usage.

### DIFF
--- a/pkg/heartbeat/heartbeat.go
+++ b/pkg/heartbeat/heartbeat.go
@@ -20,22 +20,22 @@ var remoteAddressRegex = regexp.MustCompile(`(?i)^((ssh|sftp)://)+(?P<credential
 // Heartbeat is a structure representing activity for a user on a some entity.
 type Heartbeat struct {
 	APIKey                string     `json:"-"`
-	Branch                *string    `json:"branch"`
+	Branch                *string    `json:"branch,omitempty"`
 	BranchAlternate       string     `json:"-"`
 	Category              Category   `json:"category"`
-	CursorPosition        *int       `json:"cursorpos"`
-	Dependencies          []string   `json:"dependencies"`
+	CursorPosition        *int       `json:"cursorpos,omitempty"`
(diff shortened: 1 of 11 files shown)


pkg/heartbeat/heartbeat.go
```diff
@@ -89,12 +89,12 @@ func New(
 
 // ID returns an ID generated from the heartbeat data.
 func (h Heartbeat) ID() string {
-	var branch string
+	branch := "unset"
```
alanhamlett, line 92: Is `"unset"` a special keyword when json marshaling?

alanhamlett (approved):

pkg/heartbeat/heartbeat.go
gandarez: No, it's just used for identifying a heartbeat when storing offline.

alanhamlett: I don't understand... do we swap `"unset"` with `nil` somewhere before sending to the api?

gandarez: The `ID` function returns the unique identification for heartbeat. It's only used to identify offline heartbeats like a primary key for the database. To avoid having empty strings when the field is nil I changed to "unset".

Current
```
"1592868313.541149-file-coding---/tmp/main.go-false"
```
After
```
"1592868313.541149-file-coding-unset-unset-/tmp/main.go-false"
```

alanhamlett: Both current and after are the same if only used for identifying a heartbeat right?

gandarez: yes, now we won't see empty strings for branch nor project.

alanhamlett: Nevermind, that code is running in `ID` method so it's only used there not actually setting heartbeat attributes to `"unset"`. Ignore my questions above :)

PR #828, opened 2023-02-04
Add file-experts flag

gandarez:
This PR adds new flag that acts like a command to fetch today's file experts for a given entity.


pkg/api/fileexperts.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	defer resp.Body.Close() // nolint:errcheck
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, Err{Err: fmt.Errorf("failed reading response body from %q: %s", url, err)}
+	}
+
+	switch resp.StatusCode {
+	case http.StatusOK:
```
alanhamlett: Should be:

`case http.StatusOk, http.StatusAccepted:`

pkg/fileexperts/fileexperts.go
```diff
@@ -0,0 +1,146 @@
...
+		if current != nil && other != nil {
+			break
+		}
+
+		dev := data
+
+		if data.User.IsCurrentUser {
+			current = &dev
+
+			if len(d.Data) == 1 {
```
alanhamlett: This isn't necessary, it will break in the `for` conditional above before the next iteration.

cmd/root.go
```diff
@@ -102,6 +102,7 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
 		"",
 		"(deprecated) Absolute path to file for the heartbeat."+
 			" Can also be a url, domain or app when --entity-type is not file.")
+	flags.String("file-experts", "", "Prints the top developer within a team for the given entity, then exits.")
```
alanhamlett: This should be a bool not string.

cmd/fileexperts/fileexperts.go
```diff
@@ -0,0 +1,140 @@
...
+	if v == nil {
+		return Params{}, fmt.Errorf("viper instance is nil")
+	}
+
+	if !v.IsSet("file-experts") {
+		return Params{}, fmt.Errorf("file-experts unset")
+	}
+
+	// it's a hack to get entity working with fileexpert
+	v.SetDefault("entity", vipertools.GetString(v, "file-experts"))
```
alanhamlett: We always want to require `--entity` when we have `--file-experts` and make `--file-experts` a boolean flag not a string cmd.

pkg/fileexperts/fileexperts.go
gandarez: thanks for pointing it out

cmd/root.go
gandarez: sure

alanhamlett (approved):

PR #830, opened 2023-02-08
New flag --send-diagnostics-on-errors

alanhamlett:
Also can be enabled in `~/.wakatime.cfg` with `settings.send_diagnostics_on_errors`.

### DIFF
--- a/cmd/logfile/logfile.go
+++ b/cmd/logfile/logfile.go
@@ -15,21 +15,26 @@ const defaultFile = ".wakatime.log"
 
 // Params contains log file parameters.
 type Params struct {
-	File     string
-	ToStdout bool
-	Verbose  bool
+	File              string
+	SendDiagsOnErrors bool
+	ToStdout          bool
+	Verbose           bool
 }
 
 // LoadParams loads needed data from the configuration file.
 func LoadParams(v *viper.Viper) (Params, error) {
-	var debug bool
-	if b := v.GetBool("settings.debug"); v.IsSet("settings.debug") {
-		debug = b
-	}
-
 	params := Params{
+		SendDiagsOnErrors: vipertools.FirstNonEmptyBool(
+			v,
+			"send-diagnostics-on-errors",
+			"settings.send_diagnostics_on_errors",
+		),
 		ToStdout: v.GetBool("log-to-stdout"),
-		Verbose:  v.GetBool("verbose") || debug,
+		Verbose: vipertools.FirstNonEmptyBool(
+			v,
+			"verbose",
+			"settings.debug",
+		),
 	}
 
 	logFile, ok := vipertools.FirstNonEmptyString(v, "log-file", "logfile", "settings.log_file")
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -70,61 +70,61 @@ func Run(cmd *cobra.Command, v *viper.Viper) {
 	if v.GetBool("version") {
 		log.Debugln("command: version")
 
-		RunCmd(v, logFileParams.Verbose, runVersion)
+		RunCmd(v, logFileParams.Verbose, logFileParams.SendDiagsOnErrors, runVersion)
 	}
 
 	if v.IsSet("config-read") {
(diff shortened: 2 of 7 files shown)


cmd/logfile/logfile.go
```diff
@@ -15,9 +15,10 @@ const defaultFile = ".wakatime.log"
 
 // Params contains log file parameters.
 type Params struct {
-	File     string
-	ToStdout bool
-	Verbose  bool
+	File              string
+	ToStdout          bool
+	Verbose           bool
+	SendDiagsOnErrors bool
```
gandarez: We've been ordering fields by alphabetical order.

cmd/logfile/logfile.go
```diff
@@ -27,9 +28,15 @@ func LoadParams(v *viper.Viper) (Params, error) {
...
+	if b := v.GetBool("settings.send_diagnostics_on_errors"); v.IsSet("settings.send_diagnostics_on_errors") {
+		sendDiagsOnErrors = b
+	}
+
 	params := Params{
-		ToStdout: v.GetBool("log-to-stdout"),
-		Verbose:  v.GetBool("verbose") || debug,
+		ToStdout:          v.GetBool("log-to-stdout"),
+		Verbose:           v.GetBool("verbose") || debug,
+		SendDiagsOnErrors: v.GetBool("send-diagnostics-on-errors") || sendDiagsOnErrors,
```
gandarez: Same for alphabetical order and we can replace it with `vipertools.FirstNonEmptyBool()` func, like:

```Go
SendDiagsOnErrors: vipertools.FirstNonEmptyBool("send-diagnostics-on-errors", "settings.send_diagnostics_on_errors")
```

cmd/run.go
```diff
@@ -226,26 +226,26 @@ type cmdFn func(v *viper.Viper) (int, error)
...
 		os.Exit(exitCode)
 	}
 
-	os.Exit(runCmd(v, verbose, offlinesync.Run))
+	os.Exit(runCmd(v, params, offlinesync.Run))
 }
 
 // runCmd contains the main logic of RunCmd.
-func runCmd(v *viper.Viper, verbose bool, cmd cmdFn) int {
+func runCmd(v *viper.Viper, params *logfile.Params, cmd cmdFn) int {
```
gandarez: Instead passing the whole object it's important to keep the function signature clear what's doing under the hood. It's better to pass `Verbose` and `SendDiagsOnErrors` only.

```suggestion
func runCmd(v *viper.Viper, verbose,  sendDiagsOnErrors bool, cmd cmdFn) int {
```

pkg/api/diagnostic.go
```diff
@@ -32,6 +33,7 @@ func (c *Client) SendDiagnostics(plugin string, diagnostics ...diagnostic.Diagno
 		Architecture: version.Arch,
 		CliVersion:   version.Version,
 		Plugin:       plugin,
+		IsPanic:      panic,
```
gandarez, line 34: Same here for fields ordering.

gandarez (changes requested):

cmd/logfile/logfile.go
alanhamlett: Should I do that for `debug` here too?

cmd/run.go
alanhamlett: I don't think having those params helps much... either way the reader has to go into `runCmd` to see what it's doing right?

alanhamlett: This style guide even says to use a custom type instead of passing multiple primitive variables https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

cmd/logfile/logfile.go
gandarez: Yes it makes sense to use vipertools instead.

cmd/run.go
gandarez: A struct would make it much easier to read, but this func will have only 4 input params, which is acceptable.

gandarez (approved):

PR #832, opened 2023-02-09
Fix sending diagnostics on error or panic

gandarez:


cmd/run.go
```diff
@@ -272,7 +272,7 @@ func runCmd(v *viper.Viper, verbose bool, sendDiagsOnErrors bool, cmd cmdFn) int
 		resetLogs()
 
 		if verbose && sendDiagsOnErrors {
```
alanhamlett, line 274: Oops, this line should just be `if verbose {`

cmd/run.go
```diff
@@ -255,7 +255,7 @@ func runCmd(v *viper.Viper, verbose bool, sendDiagsOnErrors bool, cmd cmdFn) int
 			resetLogs()
 
 			if verbose {
```
alanhamlett, line 257: This should be `if verbose && sendDiagsOnErrors {`

alanhamlett (approved):

PR #835, opened 2023-02-15
Add more linters and address related issues

hezhizhen:


hezhizhen:
Close https://github.com/wakatime/wakatime-cli/pull/834 with correct branch name

hezhizhen:
I need some help to pass unit tests on Windows @alanhamlett @gandarez

alanhamlett:
I re-ran that test and it passed the second time. On Windows, sometimes the GitHub Action fails to install the linter and we just re-run it.

hezhizhen:
> I re-ran that test and it passed the second time. On Windows, sometimes the GitHub Action fails to install the linter and we just re-run it.

Thanks.

alanhamlett:
LGTM, but I'll wait for @gandarez to make the final decision before merging.

.golangci.yml
```diff
@@ -15,10 +15,15 @@ linters:
     - gofmt
     - goimports
     - gosec
+    - gosimple
     - misspell
     - nestif
     - revive
     - testpackage
+    - unconvert
+    - unparam
```
gandarez: Unused params is already set using `revive:unused-parameter`

.golangci.yml
```diff
@@ -15,10 +15,15 @@ linters:
     - gofmt
     - goimports
     - gosec
+    - gosimple
```
gandarez: It's enabled by default, so don't need to explicitly set here.

.golangci.yml
```diff
@@ -15,10 +15,15 @@ linters:
...
     - misspell
     - nestif
     - revive
     - testpackage
+    - unconvert
+    - unparam
+    - unused
```
gandarez: It's enabled by default, so don't need to explicitly set here.

gandarez (changes requested):

hezhizhen:
I removed the 3 linters. @gandarez

gandarez (approved):

PR #852, opened 2023-03-01
Release v1.68.3

alanhamlett:
Changelog:
- Bump regexp2 to v1.8.1 #847
- Save heartbeats to offline db even when api key invalid #851


gandarez (approved):
The pipeline for merging into `develop` ran after few retries.

alanhamlett:
> The pipeline for merging into `develop` ran after few retries.

So strange. Guess it was an Apple temp issue.

PR #855, opened 2023-03-16
Support using git remote as project name

### LINKED ISSUE #249: Use git remote as project instead of repo folder name
Is it possible to change the plugin to get remote URL to be default project name, or maybe full path to the project name.
If so how to do it? I'm looking for a scenario where no local changes are needed by the developer.

The problem is that many projects can have the same name, so its impossible to identify them just by the project name on git.

alanhamlett:
Fixes https://github.com/wakatime/jetbrains-wakatime/issues/249

Adds a new `~/.wakatime.cfg` key:

```ini
[git]
project_from_git_remote = false
```

Defaults to `false`, but if set to `true` will use the Git remote instead of local folder as project name.

For ex: `wakatime-cli` becomes `wakatime/wakatime-cli`

### DIFF
--- a/cmd/params/params.go
+++ b/cmd/params/params.go
@@ -133,6 +133,7 @@ type (
 		BranchAlternate      string
 		MapPatterns          []project.MapPattern
 		Override             string
+		ProjectFromGitRemote bool
 		SubmodulesDisabled   []regex.Regex
 		SubmoduleMapPatterns []project.MapPattern
 	}
@@ -559,6 +560,7 @@ func loadProjectParams(v *viper.Viper) (ProjectParams, error) {
 		BranchAlternate:      vipertools.GetString(v, "alternate-branch"),
 		MapPatterns:          loadProjectMapPatterns(v, "projectmap"),
 		Override:             vipertools.GetString(v, "project"),
+		ProjectFromGitRemote: v.GetBool("git.project_from_git_remote"),
 		SubmodulesDisabled:   submodulesDisabled,
 		SubmoduleMapPatterns: loadProjectMapPatterns(v, "git_submodule_projectmap"),
 	}, nil
--- a/pkg/project/git.go
+++ b/pkg/project/git.go
@@ -13,6 +13,8 @@ import (
 type Git struct {
 	// Filepath contains the entity path.
 	Filepath string
+	// ProjectFromGitRemote when enabled uses the git remote as the project name instead of local git folder.
+	ProjectFromGitRemote bool
 	// SubmoduleDisabledPatterns will be matched against the submodule path and if matching, will skip it.
 	SubmoduleDisabledPatterns []regex.Regex
 	// SubmoduleProjectMapPatterns will be matched against the submodule path and if matching, will use the project map.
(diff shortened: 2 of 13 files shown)


cmd/params/params.go
```diff
@@ -561,6 +562,7 @@ func loadProjectParams(v *viper.Viper) (ProjectParams, error) {
 		Override:             vipertools.GetString(v, "project"),
 		SubmodulesDisabled:   submodulesDisabled,
 		SubmoduleMapPatterns: loadProjectMapPatterns(v, "git_submodule_projectmap"),
+		ProjectFromGitRemote: v.GetBool("git.project-from-git-remote"),
```
gandarez: Can we move it upwards after `Override` to keep previous ordering? Also configurations from  settings follow snake case like `git.project_from_git_remote`.

pkg/project/git.go
```diff
@@ -60,13 +62,13 @@ func (g Git) Detect() (Result, bool, error) {
 	}
 
 	// Find for .git file or directory
-	gitConfigFile, ok := FindFileOrDirectory(fp, ".git")
+	dotGitFolder, ok := FindFileOrDirectory(fp, ".git")
```
gandarez: It's either `.git` file or `.git` directory. Maybe `dotGitFolder` isn't that better than `gitConfigFile` but it's not only "dot git folder". Would you  suggest a better name for that variable?

pkg/project/git.go
```diff
@@ -246,6 +250,31 @@ func resolveCommondir(fp string) (string, bool, error) {
...
+	}
+
+	configFile := filepath.Join(dotGitFolder, "config")
+
+	remote, err := findGitRemote(configFile)
+	if err != nil {
+		log.Errorf(
+			"error finding git remote from %q: %s",
+			configFile,
+			err,
```
gandarez: There lines can be joint since them don't overcome 125 characters.

pkg/project/git.go
```diff
@@ -246,6 +250,31 @@ func resolveCommondir(fp string) (string, bool, error) {
 	return "", false, nil
 }
 
+func projectOrRemote(projectName string, g Git, dotGitFolder string) string {
+	if !g.ProjectFromGitRemote {
```
gandarez: Instead receiving the whole struct it's better to pass `ProjectFromGitRemote` as another param. This makes tests clear what should be done.

pkg/project/git.go
```diff
@@ -263,6 +292,57 @@ func findGitBranch(fp string) (string, error) {
...
+				parts = strings.SplitN(remote, ":", 2)
+				if len(parts) != 2 {
+					return "", fmt.Errorf("invalid origin url from %q: %s", fp, subline)
+				}
+
+				remote = parts[1]
+
+				remote = strings.TrimSuffix(remote, ".git")
+
+				return strings.TrimSpace(remote), nil
```
gandarez: What about
```suggestion
				remote = strings.TrimSpace(strings.TrimSuffix(parts[1], ".git")), nil
```

gandarez (changes requested):

cmd/params/params.go
alanhamlett: I thought we should use `-` because it's converted to `_` when reading from settings but dash also supports future command line params?

gandarez (approved):

Issue #558, opened 2021-09-22
Configuration files should be stored in `$XDG_CONFIG_HOME` (by default at `~/.config`)


nagromc:
**Is your feature request related to a problem? Please describe.**

I strive to have a clean home directory, but there are *many* files related to WakaTime in `~`:
- `.wakatime.bdb`
- `.wakatime.cfg`
- `.wakatime.db`
- `.wakatime.log`
- `.wakatime-internal.cfg` (since #535)

**Describe the solution you'd like to see**

[A](https://fishshell.com/docs/current/index.html#configuration) [lot](https://github.com/curl/curl/commit/3a1370ed06c8a7d2b513085a81e511917d98a6c6) [of](https://github.com/git/git/commit/ea19289bc82351b7ac20ea2fd877e2bdde97ae34) [softwares](https://github.com/alacritty/alacritty/blob/0be25c5e22ddd71a3335026b4853792512df141a/README.md#configuration) find their configuration in `$XDG_CONFIG_HOME` (by default at `~/.config`), which helps to have a less cluttered `~`.

WakaTime could use the same [convention](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) to place the configuration files such as `.wakatime.cfg` in `$XDG_CONFIG_HOME/wakatime` (which would be `~/.config/wakatime` by default), and data files such as `.wakatime.db` in `$XDG_DATA_HOME/wakatime` or `$XDG_CACHE_HOME/wakatime`.

**Describe alternatives you've considered**

Another solution would be to group them in a new directory such as `~/.wakatime`.

**Additional context**

This feature request describes the behavior in a Unix environment, but macOS and Windows have their own conventions, even if the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) leaks on them a little. I let you choose what suits the most for those platforms.

alanhamlett:
This is something we've been wanting to do for a while. Let's use:

* Unix: `$XDG_CONFIG_HOME/wakatime/`
* Mac: `$HOME/.wakatime/`
* Windows: `$HOME/.wakatime/`

Some of the editor extensions already use `$HOME/.wakatime/` for the wakatime-cli location.

nagromc:
This seems reasonable to me 👍

viggo-gascou:
Status on this? Would like to declutter my home directory as well 😅

[...]

PR #863, opened 2023-04-12
Move log file to .wakatime folder

gandarez:
This PR moves `.wakatime.log` file to `.wakatime/wakatime.log` path when **WAKATIME_HOME** env var is not set.

ref https://github.com/wakatime/wakatime-cli/issues/558#issuecomment-1502495426

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"runtime/debug"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


cmd/run.go
```diff
@@ -268,6 +268,19 @@ func runCmd(v *viper.Viper, verbose bool, sendDiagsOnErrors bool, cmd cmdFn) int
 
 			resetLogs()
 
+			// if s, ok := err.(string); ok {
```
alanhamlett: Looks like this was added by accident?

alanhamlett (changes requested):
Let's remove the commented out code and then it's good to merge.

gandarez:
> Let's remove the commented out code and then it's good to merge.

Oops! Forgot to remove local tests 👍

alanhamlett (approved):

PR #865, opened 2023-04-13
autodetect os and arch in makefile

alanhamlett:
Allows running `make build` for local dev instead of always having to specify the os and arch.


alanhamlett:
Implemented with #866.

PR #872, opened 2023-04-23
Update swift_test.go

harshsingh32:
Using `require.ElementsMatch()` function instead of `assert.Equal()` for comparing slices, will ensure that the order of elements in the slice does not matter.

### DIFF
--- a/pkg/deps/swift_test.go
+++ b/pkg/deps/swift_test.go
@@ -12,7 +12,7 @@ import (
 func TestParserSwift_Parse(t *testing.T) //TestParserSwift_Parse tests the Parse function of the ParserSwift struct.
 func TestParserSwift_Parse(t *testing.T) {
 	t.Helper()
-{
+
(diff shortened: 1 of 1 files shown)


alanhamlett (approved):

pkg/deps/swift_test.go
```diff
@@ -9,13 +9,16 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestParserSwift_Parse(t *testing.T) //TestParserSwift_Parse tests the Parse function of the ParserSwift struct.
 func TestParserSwift_Parse(t *testing.T) {
+	t.Helper()
+{
```
alanhamlett: Looks like this extra curly brace is causing a syntax error.

alanhamlett (changes requested):

gandarez:
There's no broken tests that justifies theses changes.

PR #873, opened 2023-04-23
Update version.go

harshsingh32:
Hi, @alanhamlett 
By initializing the variables to "unknown", we make it clear that the values are not set by default, will prevent any confusion that could arise from using the previous "unset" value.


gandarez:
Hi @harshsingh32 either way both have the same meaning and changing it to `unset` I don't see any difference at all.

PR #874, opened 2023-04-23
Update version.go

harshsingh32:


gandarez:
Hi, that's unnecessary and we don't define variables on function outputs.

PR #875, opened 2023-04-23
Update main.go

harshsingh32:
Includes a comment explaining the purpose of the code, a constant for the exit code, and error handling for the `os.Exit()` function call.


gandarez:
That's not a runnable code, it's only used for testing purposes, so let's keep as it is.

PR #878, opened 2023-05-06
Only log rate limited backoff message when debug mode enabled

### LINKED ISSUE #876: Only log rate limited backoff message when debug mode enabled
The backoff/rate limited error message is confusing because it's not the real problem. Let's only log that in debug mode so we can see the real error message in logs easier.

Related to https://github.com/wakatime/vscode-wakatime/issues/342.

alanhamlett:
Implements #876.

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -45,11 +45,11 @@ func Run(v *viper.Viper) (int, error) {
 				log.Errorf("failed to save heartbeats to offline queue: %s", err)
 			}
 
-			return errauth.ExitCode(), fmt.Errorf("sending heartbeat(s) failed: %s", errauth.Message())
+			return errauth.ExitCode(), fmt.Errorf("sending heartbeat(s) failed: %w", errauth)
 		}
 
 		if errwaka, ok := err.(wakaerror.Error); ok {
-			return errwaka.ExitCode(), fmt.Errorf("sending heartbeat(s) failed: %s", errwaka.Message())
+			return errwaka.ExitCode(), fmt.Errorf("sending heartbeat(s) failed: %w", errwaka)
 		}
 
 		return exitcode.ErrGeneric, fmt.Errorf(
--- a/cmd/heartbeat/heartbeat_test.go
+++ b/cmd/heartbeat/heartbeat_test.go
@@ -17,6 +17,7 @@ import (
 	cmdheartbeat "github.com/wakatime/wakatime-cli/cmd/heartbeat"
 	"github.com/wakatime/wakatime-cli/pkg/api"
 	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
+	"github.com/wakatime/wakatime-cli/pkg/ini"
 	"github.com/wakatime/wakatime-cli/pkg/log"
 	"github.com/wakatime/wakatime-cli/pkg/offline"
 	"github.com/wakatime/wakatime-cli/pkg/project"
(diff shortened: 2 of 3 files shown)


cmd/heartbeat/heartbeat.go
```diff
@@ -282,3 +262,32 @@ func setLogFields(params paramscmd.Params) {
...
+	}
+
+	if errbackoff, ok := err.(api.ErrBackoff); ok {
+		log.Debugf("sending heartbeat(s) failed: %s", errbackoff.Message())
+		return errbackoff.ExitCode(), nil
+	}
+
+	if errwaka, ok := err.(wakaerror.Error); ok {
+		return errwaka.ExitCode(), fmt.Errorf("sending heartbeat(s) failed: %s", errwaka.Message())
+	}
```
gandarez: It's not needed to filter for `api.ErrBackoff` since it implements wakaerror.Error interface. I also wouldn't move to a separate function as it moves few lines only and the `Run` method is already a small one, around 35 lines.

cmd/heartbeat/heartbeat_test.go
```diff
@@ -751,6 +752,133 @@ func TestSendHeartbeats_ErrAuth_UnsetAPIKey(t *testing.T) {
 	assert.Eventually(t, func() bool { return numCalls == 0 }, time.Second, 50*time.Millisecond)
 }
 
+func TestSendHeartbeats_ErrBackoff(t *testing.T) {
```
gandarez, line 755: I can't see where backoff is being applied in this test.

cmd/heartbeat/heartbeat_test.go
```diff
@@ -751,6 +752,133 @@ func TestSendHeartbeats_ErrAuth_UnsetAPIKey(t *testing.T) {
...
+		}
+	}()
+
+	offlineQueueFile, err := os.CreateTemp(t.TempDir(), "")
+	require.NoError(t, err)
+
+	defer offlineQueueFile.Close()
+
+	err = cmdheartbeat.SendHeartbeats(v, offlineQueueFile.Name())
+	require.Error(t, err)
```
gandarez, line 803: Maybe this assert should validate the backoff error?

cmd/run.go
```diff
@@ -331,11 +331,13 @@ func saveHeartbeats(v *viper.Viper) {
 func sendDiagnostics(v *viper.Viper, d diagnostics) error {
 	paramAPI, err := params.LoadAPIParams(v)
 	if err != nil {
-		var errauth api.ErrAuth
+		// Prevent sending diags for api key errors
```
gandarez: Jus to be consice with other comments.
```suggestion
		// Prevent sending diags for api key errors.
```

cmd/run.go
```diff
@@ -331,11 +331,13 @@ func saveHeartbeats(v *viper.Viper) {
...
+		if !errors.As(err, &api.ErrAuth{}) {
+			return fmt.Errorf("failed to load API parameters: %s", err)
+		}
 
-		// api.ErrAuth represents an error when parsing api key.
-		// In this context api key is not required to send diagnostics.
-		if !errors.As(err, &errauth) {
+		// Prevent sending diags for api connection errors
```
gandarez: same here

gandarez (changes requested):

cmd/heartbeat/heartbeat.go
alanhamlett: Linter fails when it's back in the original function.

I'm catching backoff separately here to suppress the error log by returning `nil` instead of `error`.

cmd/run.go
alanhamlett: Implemented with 1f379581fa21adef7f4e226b26ea25ddb0efd3c4.

cmd/heartbeat/heartbeat_test.go
alanhamlett: Implemented with 1f379581fa21adef7f4e226b26ea25ddb0efd3c4.

cmd/run.go
alanhamlett: Implemented with c97304448bdcae83a9895ca58f800d05b86565a3.

cmd/heartbeat/heartbeat_test.go
alanhamlett: https://github.com/wakatime/wakatime-cli/pull/878/files#diff-4e09898f30bd7131a40e051bd13906a08fd019549d103d21aaa03c3e0ff5e0a6R777

gandarez (approved):

PR #899, opened 2023-08-11
always send diags with authenticated client

gandarez:
This PR makes sending diags to always use authenticated requests.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/api/api.go
```diff
@@ -27,6 +27,9 @@ func NewClient(params paramscmd.API) (*api.Client, error) {
 // NewClientWithoutAuth initializes a new api client with all options following the
 // passed in parameters and disabled authentication.
 func NewClientWithoutAuth(params paramscmd.API) (*api.Client, error) {
+	// foce disable ssl verification despite the user's choice
```
alanhamlett: s/foce/force

alanhamlett (approved):

PR #903, opened 2023-08-14
Support ipv6 proxy address

alanhamlett:
Related to https://github.com/wakatime/vscode-wakatime/issues/361.


gandarez:
Would mind fixing the errors first? Let's also name it as `proxyIPv6Regex` or move `ipv6ProxyRegex` to be alphabetically ordered? My suggestion is to start with `proxy...`.

gandarez:
Would be great if a test case is added as well.

gandarez (approved):

PR #910, opened 2023-08-16
Try to guess the language by file content

### LINKED ISSUE #821: Language detection for files with no extension
As of the current version of wakatime cli, files which do not have an extension are tagged as "Unknown language". But often these files have shebangs inside them (eg: `#!/usr/bin/env bash`,  `#!/usr/bin/python` etc) which can be used to detect the language of the file. 

Would you consider a feature wherein the cli additionally checks for the shebang for language and then tags it as "Unknown language" if nothing is detected.

gandarez:
This PR adds a fallback that will try to guess the language by the content of a file. It also fixes `fileHead` function to read the first 512 bytes and do not fill the entire array with null characters making false-positives in some circunstances.

Fixes #821

### DIFF
--- a/pkg/language/chroma.go
+++ b/pkg/language/chroma.go
@@ -97,7 +97,48 @@ func detectChromaCustomized(filepath string) (heartbeat.Language, float32, bool)
 		return language, weight, true
 	}
 
+	// Finally, try matching by file content.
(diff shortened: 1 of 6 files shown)


alanhamlett (approved):
Perfect!

Issue #954, opened 2023-10-13
High CPU usage for wakatime-cli --today


MaximilianGaedig:
**Expected behavior (what should have happened)**:
Lower CPU usage

**Actual behavior (what went wrong)**:
Running neovim with the wakatime plugin I get a wakatime-cli with 10% CPU usage every ~15s

**Environment**:

- OS: Linux
- Platform: aarch64(M1 Pro)

**Logs**:

There are logs, but they do not get generated usually, only in some unrelated cases

alanhamlett:
Please [enable debug logging](https://github.com/wakatime/vim-wakatime#commands) then send or attach your `~/.wakatime/wakatime.log` file. I want to compare the timestamps of every logging step to see which step is taking the most time to complete, because that's most likely the part using CPU.

@gandarez when debug is enabled can we add current CPU and MEM usage to each log?
https://stackoverflow.com/questions/11356330/how-to-get-cpu-usage

MaximilianGaedig:
The only message I am getting after turning on debug each time the problem occurs: "{"caller":"cmd/today/today.go:30","func":"today.Run","level":"debug","message":"successfully fetched today for status bar","now":"2023-10-14T12:02:29+02:00","os/arch":"linux/arm64","version":"v1.85.2"}", I assume this is an expensive operation, but I don't see much of a reason why, it calls the api and just returns the result no?

Just to calrify, every 10s my statusline plugin is calling wakatime-cli --today which is causing the issue, not the official wakatime nvim plugin.

My plugin:
```lua
local Job = require 'plenary.job'
local async = require 'plenary.async'
local uv = require 'luv'

local get_wakatime_time = function()
  local tx, rx = async.control.channel.oneshot()
  local ok, job = pcall(Job.new, Job, {
    command = os.getenv 'HOME' .. '/.wakatime/wakatime-cli',
    args = { '--today' },
    on_exit = function(j, _)
      tx(j:result()[1] or '')
    end,
  })
  if not ok then
    vim.notify('Bad WakaTime call: ' .. job, 'warn')
    return ''
  end

  job:start()
  return rx()
end

---@diagnostic disable
local state = { comp_wakatime_time = '' }

-- Yield statusline value
local wakatime = function()
  local WAKATIME_UPDATE_INTERVAL = 10000

  if not Wakatime_routine_init then
    local timer = uv.new_timer()
    if timer == nil then
      return ''
    end
    -- Update wakatime every 10s
[...]

PR #955, opened 2023-10-22
Collect cpu and memory metrics

gandarez:
This PR adds new flag `--metrics` when enabled will start profiling cpu and memory to `~/.wakatime/metrics` folder. It uses the standard library to collect both metrics and save them to separate files. These files can be transformed into png images by simply calling `go tool pprof -png <file.profile>`.

Ref #954


alanhamlett (approved):

alanhamlett:
I'll modify the plugins to add `--metrics` when debug is enabled.

PR #969, opened 2023-10-26
Should always use ini date format

### LINKED ISSUE #371: Backoff rate limiter set too far in future, 2 years
The status bar counter in vscode is at 0 secs today.

I deleted the `~/.wakatime/wakatime-internal.cfg` yesterday. But the still the wakatime counter on vscode is at 0 secs today. Please find below the contents of `~/.wakatime.cfg`:

```
[settings]
 api_key = xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

 [internal]
 backoff_at      = 2022-05-04T18:17:56+05:30
 backoff_retries = 1
```

_Originally posted by @noopurphalak in https://github.com/wakatime/vscode-wakatime/issues/271#issuecomment-1780788060_

alanhamlett:
Fixes https://github.com/wakatime/vscode-wakatime/issues/371

It's not good for a function to do too many things, so I simplified `shouldBackoff` to only return one boolean.

### DIFF
--- a/pkg/backoff/backoff.go
+++ b/pkg/backoff/backoff.go
@@ -91,7 +91,7 @@ func shouldBackoff(retries int, at time.Time) (bool, bool) {
 		log.Debugf(
 			"exponential backoff tried %d times since %s, it will reset due to %d seconds limit",
 			retries,
-			at.Format(time.Stamp),
+			at.Format(ini.DateFormat),
 			resetAfter,
 		)
 
@@ -103,8 +103,8 @@ func shouldBackoff(retries int, at time.Time) (bool, bool) {
 	log.Debugf(
 		"exponential backoff tried %d times since %s, will retry at %s",
 		retries,
-		at.Format(time.Stamp),
-		at.Add(duration).Format(time.Stamp),
+		at.Format(ini.DateFormat),
+		at.Add(duration).Format(ini.DateFormat),
 	)
 
 	return time.Now().Before(at.Add(duration)), false


pkg/backoff/backoff.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	// resetAfter sets the total seconds a backoff will last.
-	resetAfter = 3600
+	// maxBackoff sets the maximum time we will rate limit before forcing to resend.
+	maxBackoff = time.Hour
```
gandarez: let's not use `time.Hour` here to be more consistent acrros the code. What about keeping 3600 seconds?
```suggestion
	maxBackoffSecs = 3600
```

pkg/backoff/backoff.go
```diff
@@ -79,35 +72,35 @@ func WithBackoff(config Config) heartbeat.HandleOption {
...
+// networking errors.
+func shouldBackoff(retries int, at time.Time) bool {
 	if retries < 1 || at.IsZero() {
-		return false, true
+		return false
 	}
 
-	next := float64(factor) * math.Pow(2, float64(retries))
-	if next > float64(resetAfter) {
+	seconds := int(float64(factor) * math.Pow(2, float64(retries)))
```
gandarez: What means "seconds"? I consider "next" better in terms of description.

pkg/backoff/backoff.go
```diff
@@ -79,35 +72,35 @@ func WithBackoff(config Config) heartbeat.HandleOption {
...
+
+	if seconds > int(maxBackoff.Seconds()) {
```
gandarez: let's change it to?
```suggestion
	if next > float64(maxBackoffSecs) {
```

pkg/backoff/backoff.go
```diff
@@ -79,35 +72,35 @@ func WithBackoff(config Config) heartbeat.HandleOption {
...
 		log.Debugf(
-			"exponential backoff tried %d times since %s, it will reset due to %d seconds limit",
+			"exponential backoff tried %d times since %s, will reset because reached %d hr max backoff",
```
gandarez: Seconds instead of hard-coded "hour"?
```suggestion
			"exponential backoff tried %d times since %s, will reset because reached %d secs max backoff",
```

pkg/backoff/backoff.go
```diff
@@ -79,35 +72,35 @@ func WithBackoff(config Config) heartbeat.HandleOption {
...
 			retries,
-			at.Format(time.Stamp),
-			resetAfter,
+			at.Format(ini.DateFormat),
+			int(maxBackoff.Hours()),
```
gandarez:
```suggestion
			maxBackoff,
```

pkg/backoff/backoff.go
```diff
@@ -79,35 +72,35 @@ func WithBackoff(config Config) heartbeat.HandleOption {
...
 
-	duration := time.Duration(next) * time.Second
-
 	log.Debugf(
 		"exponential backoff tried %d times since %s, will retry at %s",
 		retries,
-		at.Format(time.Stamp),
-		at.Add(duration).Format(time.Stamp),
+		at.Format(ini.DateFormat),
+		time.Now().Add(time.Second*time.Duration(seconds)).Format(ini.DateFormat),
```
gandarez: just for convention
```suggestion
		time.Now().Add(time.Duration(next)*time.Second).Format(ini.DateFormat),
```

pkg/backoff/backoff_test.go
```diff
@@ -38,28 +40,76 @@ func TestWithBackoff(t *testing.T) {
...
+
+	err = ini.ReadInConfig(v, tmpFile.Name())
+	require.NoError(t, err)
+
+	// make sure backoff settings were not written
+	assert.Empty(t, v.GetString("internal.backoff_at"))
+	assert.Empty(t, v.GetString("internal.backoff_retries"))
 }
 
 func TestWithBackoff_BeforeNextBackoff(t *testing.T) {
```
gandarez: This test was written to test the heartbeat won't be send during backoff, but this changes has changed it a lot and does not represent the original idea. Would be better to write another test to test backoff is written in the file. I also consider this kind of test more related to integration one.

gandarez (changes requested):
I left some comments and requested some changes to what I believe is more correct. I also didn't have time to analyze the last two new tests and I'll do it in the next loop.

pkg/backoff/backoff.go
alanhamlett: `time.Hour` conveys meaning, but `3600` doesn't. Let's start changing everywhere else in the codebase to use Durations instead of integers?

pkg/backoff/backoff.go
alanhamlett: Actually, we should use a helper function to format the `time.Duration` in human readable form, or just use the build-in formatter:

https://stackoverflow.com/questions/41335155/time-duration-to-string-2h-instead-2h0m0s

pkg/backoff/backoff.go
gandarez: `time.Hour` represents a Duration and it adds more complexity to a simple variable to control the time. Having an integer is much simpler and consumes less memory and it also don't keep the whole struct in the heap.

gandarez (approved):
👍

pkg/backoff/backoff.go
alanhamlett: Any memory side effect is so minimal it won't matter compared to code readability. I think using a duration makes it easier to read and clear vs integer.

PR #970, opened 2023-10-26
Backoff improvements

alanhamlett:
Follow-up to #969.

### DIFF
--- a/pkg/backoff/backoff.go
+++ b/pkg/backoff/backoff.go
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	// resetAfter sets the total seconds a backoff will last.
-	resetAfter = 3600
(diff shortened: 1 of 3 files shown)


pkg/backoff/backoff.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	// resetAfter sets the total seconds a backoff will last.
-	resetAfter = 3600
+	// maxBackoff sets the maximum seconds we will rate limit before forcing to resend.
+	maxBackoff = 3600
```
gandarez: Lets suffix it with "Secs" to give context?
```suggestion
	maxBackoffSecs = 3600
```

gandarez (approved):

PR #976, opened 2023-10-30
Migrate to new Apple notarytool

alanhamlett:
https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool?changes=_3_3

https://github.com/mitchellh/gon/pull/72

PR #978, opened 2023-11-01
Directly use notarytool to sign and notarize Apple binaries

alanhamlett:
After #976 didn't work, let's just directly sign and notarize the Apple binaries ourself because it's not very difficult.

This removes the dependency on gon.


gandarez (changes requested):
it also should be fixed in this PR
https://github.com/wakatime/wakatime-cli/blob/develop/bin/prepare_assets.sh#L59

alanhamlett:
> it also should be fixed in this PR https://github.com/wakatime/wakatime-cli/blob/develop/bin/prepare_assets.sh#L59

Fixed!

gandarez (approved):

PR #984, opened 2023-11-05
Prevent test side effect creating offline heartbeats

alanhamlett:
These two tests created offline heartbeats in the real `~/.wakatime.bdb` file, which would be sent to your dashboard later after running tests locally.


gandarez (approved):
Nice catch

PR #996, opened 2023-12-23
Set machine name to Gitpod when running inside Gitpod environment

alanhamlett:
https://twitter.com/Siddhant_K_code/status/1737705705476427810

Gitpod spins up a new machine often, resulting in too many Machines reported on the dashboard. This sets `Machine Name` to `Gitpod` anytime we detect the `GITPOD_WORKSPACE_ID` environment variable.


Siddhant-K-code:
Thanks @alanhamlett ❤️

PR #997, opened 2023-12-23
Support loading api key from env var

### LINKED ISSUE #761: Store `api_key` as Environment Variable
Is it possible to store the `api_key` listed in `.wakatime.cfg` file as an environmental variable instead? Perhaps something like `WAKATIME_API_KEY`. This would help by being able to create backups of my encrypted bash exports file.

alanhamlett:
When env var `WAKATIME_API_KEY` present, use it's value as the api key. This means you don't need a `~/.wakatime.cfg` file, but if the `~/.wakatime.cfg` file exists and has an `api_key` then it takes precedence over the env var.

Fixes #761.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -131,6 +131,12 @@ projects/foo = your-api-key
 ^/home/user/projects/bar(\d+)/ = your-api-key
 ```
 
+### Api Key Environment Variable
+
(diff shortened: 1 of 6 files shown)


gandarez:
I'll review it soon

PR #1022, opened 2024-01-18
Skip extra heartbeats when no data provided

gandarez:
This PR fixes reading extra heartbeat when the flag is set to `true` but no data was provided.

current log
```
failed to read extra heartbeats: failed parsing: failed to json decode from data \"\": unexpected end of JSON input
```


alanhamlett (approved):

alanhamlett:
Good catch. So this only resulted in an error logged, but the main heartbeat was still sent to the api and not lost. It's good this cleans up the `~/.wakatime/wakatime.log` so it's not noisy.

PR #1042, opened 2024-04-09
usage: clarify api_key_vault_cmd

kisaragi-hiu:
"shell command" implies it goes through a shell, whereas it is in fact only split on spaces as a command:

https://github.com/wakatime/wakatime-cli/blob/1fd560a/cmd/params/params.go#L707

So, it is a space-separated list of an executable and its arguments.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -37,7 +37,7 @@ Here's an example `$WAKATIME_HOME/.wakatime.cfg` config file with all available 
 [settings]
 debug = false
 api_key = your-api-key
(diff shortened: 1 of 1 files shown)


gandarez:
@kisaragi-hiu please follow our [guidelines](https://github.com/wakatime/wakatime-cli/blob/develop/CONTRIBUTING.md) for branching name and open another PR. Thanks.

kisaragi-hiu:
Will do, sorry.

PR #1059, opened 2024-07-09
Use new offline db filename and location

### LINKED ISSUE #1058: Move offline bdb file into wakatime folder
Just noticed the `~/.wakatime.bdb` file still lives in the user home folder, but it should be in `~/.wakatime/offline_heartbeats.bdb`.

Let's use that new location, but also check the old location when sending offline heartbeats to the API.

gandarez:
This PR moves current `~/.wakatime.bdb` from user's home folder to `~/.wakatime/offline_heartbeats.bdb`. It also syncs any pending offline heartbeats from old file and deletes it afterwards.

It reuses `offlinesync` package by calling new function `syncOfflineActivityLegacy` and logging any error as WARN. It also adds a new integration test called `TestSendHeartbeats_ExtraHeartbeats_SyncLegacyOfflineActivity` that adds extra heartbeats, sync them and also sync any pending heartbeats in the old offline queue.

closes https://github.com/wakatime/wakatime-cli/issues/1058


alanhamlett (approved):

alanhamlett:
🎉

PR #1071, opened 2024-07-23
Rate limiting by default

alanhamlett:
Sends heartbeats at most once every 2 minutes, unless:

`--heartbeat-rate-limit-seconds 0`

OR

`settings.heartbeat_rate_limit_seconds = 0`

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	apicmd "github.com/wakatime/wakatime-cli/cmd/api"
 	offlinecmd "github.com/wakatime/wakatime-cli/cmd/offline"
@@ -16,6 +17,7 @@ import (
 	"github.com/wakatime/wakatime-cli/pkg/filestats"
 	"github.com/wakatime/wakatime-cli/pkg/filter"
 	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
+	"github.com/wakatime/wakatime-cli/pkg/ini"
 	"github.com/wakatime/wakatime-cli/pkg/language"
 	_ "github.com/wakatime/wakatime-cli/pkg/lexer" // force to load all lexers
 	"github.com/wakatime/wakatime-cli/pkg/log"
@@ -75,6 +77,19 @@ func SendHeartbeats(v *viper.Viper, queueFilepath string) error {
 
 	setLogFields(params)
 	log.Debugf("params: %s", params)
+
+	if RateLimited(RateLimitParams{
+		Disabled:   params.Offline.Disabled,
+		LastSentAt: params.Offline.LastSentAt,
+		Timeout:    params.Offline.RateLimit,
+	}) {
+		if err = offlinecmd.SaveHeartbeats(v, nil, queueFilepath); err == nil {
+			return nil
+		}
+
+		// log offline db error then try to send heartbeats to API so they're not lost
+		log.Errorf("failed to save rate limited heartbeats: %s", err)
+	}
 
 	heartbeats := buildHeartbeats(params)
 
@@ -143,6 +158,10 @@ func SendHeartbeats(v *viper.Viper, queueFilepath string) error {
 		}
 	}
 
+	if err := ResetRateLimit(v); err != nil {
+		log.Errorf("failed to reset rate limit: %s", err)
+	}
+
 	return nil
 }
 
@@ -168,6 +187,48 @@ func LoadParams(v *viper.Viper) (paramscmd.Params, error) {
 		Heartbeat: heartbeatParams,
 		Offline:   paramscmd.LoadOfflineParams(v),
 	}, nil
+}
+
+// RateLimitParams contains params for the RateLimited function.
+type RateLimitParams struct {
+	Disabled   bool
+	LastSentAt time.Time
+	Timeout    time.Duration
+}
+
+// RateLimited determines if we should send heartbeats to the API or save to the offline db.
+func RateLimited(params RateLimitParams) bool {
+	if params.Disabled {
+		return false
+	}
+
+	if params.Timeout == 0 {
+		return false
+	}
+
+	if params.LastSentAt.IsZero() {
+		return false
+	}
+
+	return time.Since(params.LastSentAt) < params.Timeout
+}
+
+// ResetRateLimit updates the internal.heartbeats_last_sent_at timestamp.
+func ResetRateLimit(v *viper.Viper) error {
+	w, err := ini.NewWriter(v, ini.InternalFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to parse config file: %s", err)
+	}
+
+	keyValue := map[string]string{
+		"heartbeats_last_sent_at": time.Now().Format(ini.DateFormat),
+	}
+
+	if err := w.Write("internal", keyValue); err != nil {
+		return fmt.Errorf("failed to write to internal config file: %s", err)
+	}
+
+	return nil
 }
 
 func buildHeartbeats(params paramscmd.Params) []heartbeat.Heartbeat {
--- a/cmd/offlinesync/offlinesync.go
+++ b/cmd/offlinesync/offlinesync.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	cmdapi "github.com/wakatime/wakatime-cli/cmd/api"
+	cmdheartbeat "github.com/wakatime/wakatime-cli/cmd/heartbeat"
 	"github.com/wakatime/wakatime-cli/cmd/params"
 	"github.com/wakatime/wakatime-cli/pkg/apikey"
 	"github.com/wakatime/wakatime-cli/pkg/exitcode"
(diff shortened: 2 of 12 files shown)


gandarez:
Missing update to `USAGE.md`.

cmd/heartbeat/heartbeat.go
```diff
@@ -76,6 +78,16 @@ func SendHeartbeats(v *viper.Viper, queueFilepath string) error {
 	setLogFields(params)
 	log.Debugf("params: %s", params)
 
+	if RateLimited(params.Offline) {
+		var rateLimitedErr error
+		if rateLimitedErr = offlinecmd.SaveHeartbeats(v, nil, queueFilepath); rateLimitedErr == nil {
+			return nil
+		}
```
gandarez:
```suggestion
		if err := offlinecmd.SaveHeartbeats(v, nil, queueFilepath); err == nil {
			return nil
		}
```

cmd/heartbeat/heartbeat.go
```diff
@@ -170,6 +182,23 @@ func LoadParams(v *viper.Viper) (paramscmd.Params, error) {
 	}, nil
 }
 
+// RateLimited determines if we should send heartbeats to the API or save to the offline db.
+func RateLimited(offline params.Offline) bool {
```
gandarez: Instead passing the whole struct, let's keep it simple by only passing what's used in the function. It's also missing tests for this function.
```suggestion
func RateLimited(offlineDisabled bool, offlineRateLimit int, offlineLastSentAt time) bool {
```

alanhamlett:
> Missing update to `USAGE.md`.

Fixed!

cmd/offlinesync/offlinesync.go
```diff
@@ -16,8 +17,30 @@ import (
...
+
+	return Run(v, paramOffline)
+}
+
+// RunWithRateLimiting executes sync-offline-activity command with rate limiting enabled.
+func RunWithRateLimiting(v *viper.Viper) (int, error) {
+	paramOffline := params.LoadOfflineParams(v)
+
+	if cmdheartbeat.RateLimited(paramOffline) {
+		return exitcode.Success, nil
```
gandarez, line 35: Maybe adding a debug message to say it's rate limited and heartbeats weren't synced at this time?

cmd/offlinesync/offlinesync.go
```diff
@@ -16,8 +17,30 @@ import (
...
+	}
+
+	return Run(v, paramOffline)
+}
+
+// Run executes sync-offline-activity command.
+func Run(v *viper.Viper, paramOffline params.Offline) (int, error) {
```
gandarez: The same idea here, to avoid passing the whole struct and simplify arguments.

cmd/offlinesync/offlinesync.go
```diff
@@ -53,13 +76,11 @@ func Run(v *viper.Viper) (int, error) {
 
 // syncOfflineActivityLegacy syncs the old offline activity by sending heartbeats
 // from the legacy offline queue to the WakaTime API.
-func syncOfflineActivityLegacy(v *viper.Viper, queueFilepath string) error {
+func syncOfflineActivityLegacy(v *viper.Viper, queueFilepath string, paramOffline params.Offline) error {
 	if queueFilepath == "" {
 		return nil
 	}
 
-	paramOffline := params.LoadOfflineParams(v)
```
gandarez: Can we keep loading params offline here?

cmd/offlinesync/offlinesync.go
```diff
@@ -96,9 +117,7 @@ func syncOfflineActivityLegacy(v *viper.Viper, queueFilepath string) error {
 
 // SyncOfflineActivity syncs offline activity by sending heartbeats
 // from the offline queue to the WakaTime API.
-func SyncOfflineActivity(v *viper.Viper, queueFilepath string) error {
-	paramOffline := params.LoadOfflineParams(v)
```
gandarez, line 100: The same here, can we keep loading params offline here?

gandarez (changes requested):

gandarez:
Tests are failing.

cmd/heartbeat/heartbeat.go
alanhamlett: We need the `err` in the line below for logging, but that `err` is actually from:
https://github.com/wakatime/wakatime-cli/pull/1071/files#diff-a42970af11c59a3cbee6b9abd656868ab6508da9003c66d042bcba1f0514811aR72

Because of scoping, so it's not actually the `err` we're looking for.

cmd/offlinesync/offlinesync.go
alanhamlett: That's doing unnecessary work, because we've already loaded offline params in both the parent functions.

cmd/heartbeat/heartbeat.go
alanhamlett: I actually like passing the struct here because it's called from two different files where we have the struct already, and we don't really care about the details of the individual params, we mainly care about the return value from `RateLimited`. It makes the code read as english easier:

`if rate limited` vs `if rate limited given bool, int, time`

alanhamlett:
> Tests are failing.

Fixed tests!

cmd/heartbeat/heartbeat.go
gandarez: My suggestion makes `err` scoped to `if` statement. @alanhamlett

pkg/offline/offline.go
```diff
@@ -33,6 +33,9 @@ const (
 	// SyncMaxDefault is the default maximum number of heartbeats from the
 	// offline queue, which will be synced upon sending heartbeats to the API.
 	SyncMaxDefault = 1000
+	// RateLimitDefaultSeconds is the default seconds between sending heartbeats
+	// to the API. If not enough time has passed, heartbeats are saved to the offline queue.
+	RateLimitDefaultSeconds = 120
```
gandarez: Could we make it alphabetically ordered as well?

cmd/offlinesync/offlinesync.go
gandarez: It's using pointers so there's no overhead calling it again. It's also recommended to not pass unused params to the functions making it harder to test and guessing what's used or not.

cmd/heartbeat/heartbeat.go
```diff
@@ -170,6 +181,23 @@ func LoadParams(v *viper.Viper) (paramscmd.Params, error) {
 	}, nil
 }
 
+// RateLimited determines if we should send heartbeats to the API or save to the offline db.
+func RateLimited(offline paramscmd.Offline) bool {
```
gandarez: missing unit test?

cmd/params/params.go
```diff
@@ -647,18 +649,41 @@ func LoadOfflineParams(v *viper.Viper) Offline {
 		disabled = !b
 	}
 
+	rateLimit := v.GetInt("heartbeat-rate-limit-seconds")
```
gandarez: missing unit test?

cmd/params/params.go
```diff
@@ -647,18 +649,41 @@ func LoadOfflineParams(v *viper.Viper) Offline {
...
 	syncMax := v.GetInt("sync-offline-activity")
 	if syncMax < 0 {
 		log.Warnf("argument --sync-offline-activity must be zero or a positive integer number, got %d", syncMax)
 
 		syncMax = 0
 	}
 
+	var lastSentAt time.Time
+
+	lastSentAtStr := vipertools.GetString(v, "internal.heartbeats_last_sent_at")
```
gandarez, line 668: missing unit test?

gandarez (changes requested):

cmd/heartbeat/heartbeat.go
alanhamlett: Yea but we need that err outside the if statement 2 lines down.

alanhamlett: I'll modify the code to be more clear.

alanhamlett: afb0fd295619dde0ced2ba615b79ec8238781b52

cmd/heartbeat/heartbeat.go
gandarez: Instead we could create a struct to hold this params like we've done [here](https://github.com/wakatime/wakatime-cli/blob/develop/cmd/heartbeat/heartbeat.go#L234-L273).

cmd/heartbeat/heartbeat.go
```diff
@@ -76,6 +77,15 @@ func SendHeartbeats(v *viper.Viper, queueFilepath string) error {
 	setLogFields(params)
 	log.Debugf("params: %s", params)
 
+	if RateLimited(params.Offline) {
+		if err := offlinecmd.SaveHeartbeats(v, nil, queueFilepath); err != nil {
```
gandarez: linter is warning with
```
early-return: if c { ... } else { ... return } can be simplified to if !c { ... return } ... (move short variable declaration to its own line if necessary) (revive)
```

alanhamlett: That's how I had it previously

alanhamlett: I'll change it back, but we need the variable to not scoped to the `if` block.

cmd/heartbeat/heartbeat.go
alanhamlett: Done.

PR #1076, opened 2024-07-24
Read extra heartbeats once

gandarez:
This PR prevents reading extra heartbeats more than once by setting `extra-heartbeats` flag to false when rate limited. It also includes an integration test to assert this behavior.

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -70,7 +70,7 @@ func Run(v *viper.Viper) (int, error) {
 // heartbeats from the offline queue, if available and offline sync is not
 // explicitly disabled.
 func SendHeartbeats(v *viper.Viper, queueFilepath string) error {
-	params, err := LoadParams(v)
+	params, err := loadParams(v)
 	if err != nil {
 		return fmt.Errorf("failed to load command parameters: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


cmd/params/params.go
```diff
@@ -750,29 +752,20 @@ func readAPIKeyFromCommand(cmdStr string) (string, error) {
 	return strings.TrimSpace(string(out)), nil
 }
 
-var extraHeartbeatsCache *[]heartbeat.Heartbeat // nolint:gochecknoglobals
```
alanhamlett, line 753: I still want to use a semaphore to be sure it never reads stdin twice. It can't hurt.

cmd/heartbeat/heartbeat.go
```diff
@@ -83,7 +83,11 @@ func SendHeartbeats(v *viper.Viper, queueFilepath string) error {
 		LastSentAt: params.Offline.LastSentAt,
 		Timeout:    params.Offline.RateLimit,
 	}) {
-		if err = offlinecmd.SaveHeartbeats(v, nil, queueFilepath); err == nil {
+		// it prevents reading extra heartbeats again from stdin when we're rate limited.
+		// Otherwise, it would fail to read.
+		v.Set("extra-heartbeats", false)
```
alanhamlett, line 88: I don't like this... it seems brittle.

PR #1077, opened 2024-07-25
Cleanup tests and remove unused pointer

alanhamlett:


cmd/heartbeat/heartbeat_test.go
```diff
@@ -35,6 +35,7 @@ import (
 )
 
 func TestSendHeartbeats(t *testing.T) {
+	resetSingletons(t)
```
gandarez: nit: break line
```suggestion
	resetSingletons(t)
	
```

gandarez (approved):

PR #1106, opened 2024-10-08
Support nested heartbeat response status in 2xx range

alanhamlett:
The HTTP response needs 201 or 202 because some proxies or other network configurations always return 200 status even for failed requests that don't reach the WakaTime API. Because the WakaTime API never responds to POST heartbeats with 200 status, we want to treat 200 status as an error and save the heartbeats to the local offline db.

But for the nested JSON array in the response content we can accept any 2xx status code.

### DIFF
--- a/pkg/api/heartbeat.go
+++ b/pkg/api/heartbeat.go
@@ -135,7 +135,7 @@ func parseHeartbeatResponse(data []json.RawMessage) (heartbeat.Result, error) {
 		return heartbeat.Result{}, fmt.Errorf("failed to parse json status: %s", err)
 	}
 
-	if result.Status >= http.StatusBadRequest {
+	if result.Status < http.StatusOK || result.Status > 299 {
 		resultErrors, err := parseHeartbeatResponseError(data[0])
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/api/heartbeat.go
```diff
@@ -135,7 +135,7 @@ func parseHeartbeatResponse(data []json.RawMessage) (heartbeat.Result, error) {
 		return heartbeat.Result{}, fmt.Errorf("failed to parse json status: %s", err)
 	}
 
-	if result.Status >= http.StatusBadRequest {
+	if result.Status < 200 || result.Status > 299 {
```
gandarez: Can we use the builtin constant? @alanhamlett

alanhamlett: I wanted to, but there's no constant for `299` and the constant for `300` (`http.StatusMultipleChoices`) doesn't sound like an error so it's more confusing than just using `299`.

gandarez: We can use at least for Status Ok?

alanhamlett: Done!

gandarez (approved):

PR #1119, opened 2024-11-08
Replace logrus by uber-go/zap

gandarez:
This PR replaces the `logrus` logger by `uber-go/zap`. According to the benchmarks zap is 1677% faster than logrus.


alanhamlett (approved):
🎉

PR #1146, opened 2024-12-28
Save config-based logger to context for shared use

leekib:
The logger configured with setting file was not stored in the context, making it inaccessible across the application.
As a result, settings like `debug = true` in `wakatime.cfg` were not consistently applied since the logger was not shared via the context.

This PR adds a line to ensure the logger is stored in the context.


alanhamlett:
This was just merged as part of #1150.

alanhamlett:
Thanks @leekib!

PR #1153, opened 2025-01-06
Fix rate limit custom number

### LINKED ISSUE #1147: --heartbeat-rate-limit-seconds flag is not properly configured in the config file
I’m using the VSCode extension, and even though I set `heartbeat_rate_limit_seconds = 10` in the `wakatime.cfg` file, it seems that the actual `heartbeat_rate_limit_seconds` is being set to the default value of 120.  

Looking at the code, it appears that the `heartbeat_rate_limit_seconds` value from `wakatime.cfg` is only read if the flag is explicitly set to 0. This behavior feels more like a bug than an intended design.  

Additionally, even though I’ve set `debug = true`, I’m unable to see any debug logs. I’ve already made a simple fix for this issue and submitted a PR. Please let me know if I’ve missed anything.

I recently discovered WakaTime and thought it would be fun to show, in near real-time, what code I’m working on through my blog.  So I hope this issue gets resolved soon!

**Environment**:

- OS: mac os 15.1.1(24B91)
- Platform: arm64

**Logs**:

gandarez:
This PR fixes setting custom number for `heartbeat-rate-limit-seconds`. Currently any value informed from config is completely ignored as the default value for the flag is 120. The easiest way was to make the default value zero as `vipertools.FirstNonEmptyInt()` only returns valid values that are not equal to zero and making it parsing correctly when set by flag or config file. @alanhamlett the drawback here is to instead of setting rate limit to zero it needs to be set to -1 (_or any negative number_) to get disabled.

Fixes #1147

### DIFF
--- a/cmd/params/params_test.go
+++ b/cmd/params/params_test.go
@@ -20,14 +20,15 @@ import (
 	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
 	inipkg "github.com/wakatime/wakatime-cli/pkg/ini"
 	"github.com/wakatime/wakatime-cli/pkg/log"
+	"github.com/wakatime/wakatime-cli/pkg/offline"
 	"github.com/wakatime/wakatime-cli/pkg/output"
 	"github.com/wakatime/wakatime-cli/pkg/project"
 	"github.com/wakatime/wakatime-cli/pkg/regex"
-	"gopkg.in/ini.v1"
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/ini.v1"
 )
 
 func TestLoadHeartbeatParams_AlternateProject(t *testing.T) {
@@ -1832,7 +1833,7 @@ func TestLoadOfflineParams_RateLimit_FromConfig(t *testing.T) {
 
 func TestLoadOfflineParams_RateLimit_Zero(t *testing.T) {
 	v := viper.New()
-	v.Set("heartbeat-rate-limit-seconds", "0")
+	v.Set("heartbeat-rate-limit-seconds", 0)
 
 	params := cmdparams.LoadOfflineParams(context.Background(), v)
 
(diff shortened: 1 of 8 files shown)


USAGE.md
```diff
@@ -91,7 +91,7 @@ some/submodule/name = new project name
 | api_key                        | Your wakatime api key. | _string_ | |
 | api_key_vault_cmd              | A command to get your api key, perhaps from some sort of secure vault. Actually a space-separated list of an executable and its arguments. Executables in PATH can be referred to by their basenames. Shell syntax not supported. | _string_ | |
 | api_url                        | The WakaTime API base url. | _string_ | <https://api.wakatime.com/api/v1> |
-| heartbeat_rate_limit_seconds   | Rate limit sending heartbeats to the API once per duration. Set to 0 to disable rate limiting. | _int_ | `120` |
+| heartbeat_rate_limit_seconds   | Rate limit sending heartbeats to the API once per duration. Set to -1 to disable rate limiting. | _int_ | `120` |
```
leekib: It seems there's a same issue with the `timeout` as well.

cmd/params/params_test.go
```diff
@@ -1863,7 +1864,7 @@ func TestLoadOfflineParams_RateLimit_NonIntegerValue(t *testing.T) {
 
 	params := cmdparams.LoadOfflineParams(context.Background(), v)
 
-	assert.Zero(t, params.RateLimit)
+	assert.Equal(t, time.Duration(offline.RateLimitDefaultSeconds)*time.Second, params.RateLimit)
```
gandarez, line 1867: Invalid values when set, should default to `offline.RateLimitDefaultSeconds`.

gandarez:
Closing in favor of https://github.com/wakatime/wakatime-cli/pull/1154

USAGE.md
gandarez: @leekib thanks for pointing it out, I fixed it with https://github.com/wakatime/wakatime-cli/pull/1154

PR #1154, opened 2025-01-06
Fix unhandled default params

### LINKED ISSUE #1147: --heartbeat-rate-limit-seconds flag is not properly configured in the config file

gandarez:
This PR fixes unhandled default params when flag/config is not set. Currently any value informed from flag/config is completely ignored for `rate limit` and `timeout`.  The `vipertools.FirstNonEmptyInt()` function now checks if value is set and accepts zero as valid one.

Some tests have been added to ensure the right behavior and increase coverage.

Fixes #1147

### DIFF
--- a/cmd/params/params.go
+++ b/cmd/params/params.go
@@ -22,6 +22,7 @@ import (
 	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
 	"github.com/wakatime/wakatime-cli/pkg/ini"
 	"github.com/wakatime/wakatime-cli/pkg/log"
+	"github.com/wakatime/wakatime-cli/pkg/offline"
 	"github.com/wakatime/wakatime-cli/pkg/output"
 	"github.com/wakatime/wakatime-cli/pkg/project"
 	"github.com/wakatime/wakatime-cli/pkg/regex"
@@ -296,10 +297,10 @@ func LoadAPIParams(ctx context.Context, v *viper.Viper) (API, error) {
 		}
 	}
 
-	var timeout time.Duration
+	timeout := api.DefaultTimeoutSecs
 
 	if timeoutSecs, ok := vipertools.FirstNonEmptyInt(v, "timeout", "settings.timeout"); ok {
-		timeout = time.Duration(timeoutSecs) * time.Second
+		timeout = timeoutSecs
 	}
 
 	return API{
@@ -312,7 +313,7 @@ func LoadAPIParams(ctx context.Context, v *viper.Viper) (API, error) {
 		Plugin:           vipertools.GetString(v, "plugin"),
 		ProxyURL:         proxyURL,
 		SSLCertFilepath:  sslCertFilepath,
-		Timeout:          timeout,
+		Timeout:          time.Duration(timeout) * time.Second,
 		URL:              apiURL.String(),
 	}, nil
 }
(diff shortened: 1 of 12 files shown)


cmd/params/params_test.go
```diff
@@ -1863,7 +1908,7 @@ func TestLoadOfflineParams_RateLimit_NonIntegerValue(t *testing.T) {
 
 	params := cmdparams.LoadOfflineParams(context.Background(), v)
 
-	assert.Zero(t, params.RateLimit)
+	assert.Equal(t, time.Duration(offline.RateLimitDefaultSeconds)*time.Second, params.RateLimit)
```
gandarez, line 1911: Invalid values when set, should default to `offline.RateLimitDefaultSeconds`.

cmd/params/params.go
```diff
@@ -1100,7 +1109,7 @@ func (p Offline) String() string {
 	}
 
 	return fmt.Sprintf(
-		"disabled: %t, last sent at: '%s', print max: %d, num rate limit: %d, num sync max: %d",
+		"disabled: %t, last sent at: '%s', print max: %d, rate limit: %s, num sync max: %d",
```
gandarez, line 1113: Correctly print `time.Duration` format.

cmd/params/params_test.go
```diff
@@ -2255,19 +2303,15 @@ func TestLoadAPIParams_BackoffAtErr(t *testing.T) {
...
 
-	assert.Equal(t, cmdparams.API{
-		BackoffAt:      time.Time{},
-		BackoffRetries: 2,
-		Key:            "00000000-0000-4000-8000-000000000000",
-		URL:            "https://api.wakatime.com/api/v1",
-		Hostname:       "my-computer",
-	}, params)
+	assert.Equal(t, 2, params.BackoffRetries)
+	assert.Empty(t, params.BackoffAt)
```
gandarez, line 2314: Simplify the testing scope to what matters.

pkg/vipertools/vipertools.go
```diff
@@ -31,9 +32,20 @@ func FirstNonEmptyInt(v *viper.Viper, keys ...string) (int, bool) {
...
+			continue
+		}
+
+		// Zero means a valid value when set, so it needs to use generic function and later cast it to int
+		value := v.Get(key)
+
+		// If the value is not an int, it will continue to find the next non-empty key
+		parsed, err := cast.ToIntE(value)
+		if err != nil {
+			continue
```
gandarez, line 45: if casting to int fails, then continue to next non-empty value if exists.

pkg/vipertools/vipertools_test.go
```diff
@@ -66,7 +66,7 @@ func TestFirstNonEmptyInt_EmptyInt(t *testing.T) {
 	v := viper.New()
 	v.Set("first", 0)
 	_, ok := vipertools.FirstNonEmptyInt(v, "first")
-	assert.False(t, ok)
+	assert.True(t, ok)
```
gandarez, line 69: Zero values are now considered a valid ones when set.

main_test.go
```diff
@@ -249,6 +249,88 @@ func TestSendHeartbeats_SecondaryApiKey(t *testing.T) {
 	assert.Eventually(t, func() bool { return numCalls == 1 }, time.Second, 50*time.Millisecond)
 }
 
+func TestSendHeartbeats_Timeout(t *testing.T) {
```
gandarez, line 252: New integration test to make sure it fails by timeout other than generic errors.

alanhamlett (approved):

PR #1155, opened 2025-01-06
Accept empty values when set in vipertools

### LINKED ISSUE #822: Precedence flag won't take effect when set to default value
Consider the following parameters (_or any that accepts either from flag or config file_)
* exclude-unknown-project
* settings.exclude_unknown_project
If settings is `true` and you would like to temporary disable when calling the cli by passing `exclude-unknown-project false` it won’t take effect.

gandarez:
This PR will refactor `vipertools.FirstNonEmptyBool` and `vipertools.FirstNonEmptyString` to accept empty values for each type (_false/empty string_).

Closes #822

### DIFF
--- a/cmd/params/params_test.go
+++ b/cmd/params/params_test.go
@@ -144,7 +144,7 @@ func TestLoadHeartbeatParams_CursorPosition_Unset(t *testing.T) {
 	assert.Nil(t, params.CursorPosition)
 }
 
-func TestLoadHeartbeatParams_Entity_EntityFlagTakesPrecedence(t *testing.T) {
+func TestLoadHeartbeatParams_Entity_FlagTakesPrecedence(t *testing.T) {
 	v := viper.New()
 	v.Set("entity", "/path/to/file")
 	v.Set("file", "ignored")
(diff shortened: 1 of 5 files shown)


cmd/params/params_test.go
```diff
@@ -925,6 +925,18 @@ func TestLoadHeartbeatParams_Filter_ExcludeUnknownProject_FromConfig(t *testing.
...
+func TestLoadHeartbeatParams_Filter_ExcludeUnknownProject_FlagTakesPrecedence(t *testing.T) {
+	v := viper.New()
+	v.Set("entity", "/path/to/file")
+	v.Set("exclude-unknown-project", false)
+	v.Set("settings.exclude_unknown_project", true)
+
+	params, err := cmdparams.LoadHeartbeatParams(context.Background(), v)
+	require.NoError(t, err)
+
+	assert.False(t, params.Filter.ExcludeUnknownProject)
```
gandarez, line 936: It should be false despite it's set in config file, and flag always takes precedence.

alanhamlett (approved):

PR #1156, opened 2025-01-06
New config hide_dependencies

alanhamlett:
Adds cli param `--hide-dependencies` and `~/.wakatime.cfg` config `settings.hide_dependencies`. When `true` or a list of regex patterns that matches the current entity file path, imports/libraries/dependencies aren't sent to the api.


gandarez:
@alanhamlett we could also make it by project and maybe branch? If certain restricted projects needs to get dependencies hidden or some branches, like `master`/`main`.

alanhamlett:
> @alanhamlett we could also make it by project and maybe branch? If certain restricted projects needs to get dependencies hidden or some branches, like `master`/`main`.

Just by entity file path, because that equals project normally. I don't think anyone needs to hide dependencies for a branch in a project while sending for another branch. Even with a monorepo they would want to hide dependencies based on folder path of the files being edited.

gandarez:
Fair enough, let's get it released and collect some feedbacks.

PR #1166, opened 2025-04-02
Bump Go to v1.24.2

alanhamlett:


alanhamlett:
Implemented with #1167.

PR #1182, opened 2025-06-28
[WIP] Add project-level configuration support with .wakatime files

### LINKED ISSUE #884: Add Support for project-level Config Files
Basically, `project_api_key`, but with `api_url`. For example, `project_api_url`.

The above is the simple & quick solution for one bigger problem. The ideal solution for all problems on the other hand would be each project having its own optional configuration file, overwriting settings from the global file. So, if the project file has a different `api_url` only that one gets overriden.

We already may use a `.wakatime-project`, so why not actually making it work (or simply adding a normal `.wakatime` in the project directory)?

iceice666:
# Summary

Implements project-level configuration support as requested in [#884](https://github.com/wakatime/wakatime-cli/issues/884). This allows projects to override global configuration settings using `.wakatime` files in their directory structure.

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -172,12 +172,14 @@ func LoadParams(ctx context.Context, v *viper.Viper) (paramscmd.Params, error) {
 		return paramscmd.Params{}, errors.New("viper instance unset")
 	}
 
-	apiParams, err := paramscmd.LoadAPIParams(ctx, v)
+	// Load API params with project config support
+	apiParams, err := paramscmd.LoadAPIParamsWithProjectConfig(ctx, v)
 	if err != nil {
 		return paramscmd.Params{}, fmt.Errorf("failed to load API parameters: %w", err)
 	}
 
-	heartbeatParams, err := paramscmd.LoadHeartbeatParams(ctx, v)
+	// Load heartbeat params with project config support
+	heartbeatParams, err := paramscmd.LoadHeartbeatParamsWithProjectConfig(ctx, v)
 	if err != nil {
 		return paramscmd.Params{}, fmt.Errorf("failed to load heartbeat params: %s", err)
 	}
--- a/cmd/params/params.go
+++ b/cmd/params/params.go
@@ -168,6 +168,21 @@ type (
 	}
 )
 
+// LoadAPIParamsWithProjectConfig loads API params from viper.Viper instance
+// with project-level configuration merged in. Returns ErrAuth if failed to retrieve api key.
+func LoadAPIParamsWithProjectConfig(ctx context.Context, v *viper.Viper) (API, error) {
+	// Get the entity first to determine the project config
+	entity := vipertools.FirstNonEmptyString(v, "entity", "file")
+	if entity != "" {
+		// Load project-specific configuration if available
+		mergedViper := project.LoadProjectConfig(ctx, v, entity)
+		return LoadAPIParams(ctx, mergedViper)
+	}
+
+	// Fall back to regular loading if no entity is provided
+	return LoadAPIParams(ctx, v)
+}
+
 // LoadAPIParams loads API params from viper.Viper instance. Returns ErrAuth
 // if failed to retrieve api key.
 func LoadAPIParams(ctx context.Context, v *viper.Viper) (API, error) {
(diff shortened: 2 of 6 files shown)


alanhamlett:
What AI tool wrote your PR and code?

cmd/heartbeat/heartbeat.go
```diff
@@ -172,12 +172,14 @@ func LoadParams(ctx context.Context, v *viper.Viper) (paramscmd.Params, error) {
...
-	apiParams, err := paramscmd.LoadAPIParams(ctx, v)
+	// Load API params with project config support
+	apiParams, err := paramscmd.LoadAPIParamsWithProjectConfig(ctx, v)
 	if err != nil {
 		return paramscmd.Params{}, fmt.Errorf("failed to load API parameters: %w", err)
 	}
 
-	heartbeatParams, err := paramscmd.LoadHeartbeatParams(ctx, v)
+	// Load heartbeat params with project config support
+	heartbeatParams, err := paramscmd.LoadHeartbeatParamsWithProjectConfig(ctx, v)
```
alanhamlett, line 182: We don't need to rename these methods. Keep the names the same. Comments also aren't necessary here.

cmd/params/params.go
```diff
@@ -168,6 +168,21 @@ type (
 	}
 )
 
+// LoadAPIParamsWithProjectConfig loads API params from viper.Viper instance
+// with project-level configuration merged in. Returns ErrAuth if failed to retrieve api key.
+func LoadAPIParamsWithProjectConfig(ctx context.Context, v *viper.Viper) (API, error) {
+	// Get the entity first to determine the project config
+	entity := vipertools.FirstNonEmptyString(v, "entity", "file")
```
alanhamlett, line 175: This is buggy. It doesn't use the same pattern for getting entity as the original code, and it doesn't account for entities in extra heartbeats.

alanhamlett (changes requested):
The approach doesn't account for extra heartbeats, reading `.wakatime` project config files needs to happen at a different layer.

In general, please don't submit AI generated code without heavily modifying it and vetting it yourself. Remove all AI-generated inline code comments, they distract from what the code is actually doing.

cmd/params/params.go
alanhamlett: Also it forgets about `--project-folder` param and `project_folder` extra heartbeat json.

alanhamlett: Also, [there's](https://github.com/wakatime/wakatime-cli/blob/65ef62fb3a900cefd617270bf539b13880596978/cmd/today/today.go#L41) [many](https://github.com/wakatime/wakatime-cli/blob/65ef62fb3a900cefd617270bf539b13880596978/cmd/fileexperts/fileexperts.go#L91) [other](https://github.com/wakatime/wakatime-cli/blob/65ef62fb3a900cefd617270bf539b13880596978/cmd/offline/offline.go#L61) [places](https://github.com/wakatime/wakatime-cli/blob/65ef62fb3a900cefd617270bf539b13880596978/cmd/offlinesync/offlinesync.go#L98) [where](https://github.com/wakatime/wakatime-cli/blob/65ef62fb3a900cefd617270bf539b13880596978/cmd/todaygoal/todaygoal.go#L79) `LoadAPIParams` is called and shouldn't forget project configs.

Another consideration is what happens when coding without internet... we save heartbeats to the offline db without the api key currently. We have to make sure the project folder's `.wakatime` file is still readable even if the entity file no longer exists when sending the offline heartbeat. For ex:

* Entity is `/home/user/project/pkg/project/config.go`
* Project config is at `/home/user/project/.wakatime`
* Heartbeat saved to offline db
* Folder `/home/user/project/pkg/project/` is deleted
* Send offline heartbeat to API, but walking the directory path results in a filesystem error now so we never find the project config at `/home/user/project/.wakatime` or even worse, we fail to send the heartbeat

alanhamlett: Looking at `project.FindFileOrDirectory` it doesn't error and continues to the parent folder when a path doesn't exist on the filesystem, so we don't need to worry about the entity existing when sending offline heartbeats. But the first point in [my comment](https://github.com/wakatime/wakatime-cli/pull/1182#discussion_r2173212975) stands about `LoadAPIParams` calls need to all be consistent and use project config everywhere params are loaded.

pkg/project/config.go
```diff
@@ -0,0 +1,109 @@
...
+
+	logger := log.Extract(ctx)
+	logger.Debugf("wakatime config file found at: %s", fp)
+
+	// Load the INI file
+	cfg, err := ini.LoadSources(ini.LoadOptions{
+		AllowPythonMultilineValues: true,
+		SkipUnrecognizableLines:    true,
+	}, fp)
+	if err != nil {
```
alanhamlett, line 49: My guess is this will always error when it finds the `~/.wakatime` folder and tries to read the folder as a project config file, effectively breaking the plugins for all users who don't have a `.wakatime` project config.

iceice666:
I'm using AI tool (Claude Code) to help me understand entire codebase and ask it to optimize my implementation.

I will rewrite entire PR.

Apologize for any inconvenience.

alanhamlett:
It's no inconvenience.

To do this correctly we need to associate params with individual heartbeats in an array, passing that array into a sender function that sends all the heartbeats and reads the project config if available for each heartbeat based on that heartbeat's entity and project_folder. The sender function should take an array of viper.Viper and heartbeat.Heartbeat as input, read any project config and merge into a copy of the existing viper.Viper, then call LoadParams for each heartbeat to build that heartbeat's params including project configs.

It's going to be a large refactor of the existing code structure.

iceice666:
Thanks for the detailed explanation of what would be needed to implement this correctly.

Given that this requires a large refactor of the existing code structure and I'm not familiar enough with the codebase to undertake such changes, I'll close this PR.

I appreciate your time reviewing this and the guidance on the proper implementation approach.

PR #1183, opened 2025-07-01
Implement project-level config

### LINKED ISSUE #884: Add Support for project-level Config Files

gandarez:
This PR implements the detection for project level configuration. It traverses the entity's folder looking for `.wakatime` file and if found merges the configuration with the global. For each heartbeat it tries to find specific file. 

- Using the new package `handler` to handle the preprocessing options which includes all filters applied in a heartbeat before actually calling the API. Prior this change wasn't possible to read specific configuration file per heartbeat because the processors (options) were created using the global configuration only. Bulk send is still working.
- All `initHandleOptions` functions have been refactored to return an array of Preprocessor type instead of HandleOption. This is needed to dynamically load the Option using the config passed in.
- Added new integration test `TestSendHeartbeats_ExtraHeartbeats_ProjectConfigFile` which sends two heartbeats to the cli where the latter has a custom `.wakatime` project file in the source tree.

closes #884


alanhamlett (approved):

alanhamlett:
Awesome!

theAkito:
Thank you very much! Will try!

PR #1196, opened 2025-07-15
Add handler package

gandarez:
This PR creates a new package `handler` in cmd to handle the preprocessing options which includes all filters applied in a heartbeat before actually calling the API.

### DIFF
--- a/cmd/handler/handler.go
+++ b/cmd/handler/handler.go
@@ -1,0 +1,131 @@
+package handler
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/wakatime/wakatime-cli/pkg/file"
+	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
+	"github.com/wakatime/wakatime-cli/pkg/ini"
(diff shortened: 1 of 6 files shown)


alanhamlett (approved):

cmd/handler/handler.go
```diff
@@ -0,0 +1,131 @@
...
+		return paramspkg.Params{}, false, nil
+	}
+
+	vproj, err := vipertools.New()
+	if err != nil {
+		return paramspkg.Params{}, false, fmt.Errorf("failed to create viper instance: %w", err)
+	}
+
+	// copy only settings from the main viper instance to the project-level viper instance
+	vipertools.CopyOnlySettings(v, vproj)
```
alanhamlett, line 103: Does this mean if we find a project config, it removes project_map, project_api_key, and other non-settings sections from the main config?

gandarez: partially, it does skip only the ones present in this [list](https://github.com/wakatime/wakatime-cli/blob/develop/pkg/vipertools/vipertools.go#L30).

PR #1204, opened 2025-07-22
Improve category parsing

gandarez:
This PR improves how category is parsing. It does the following:
* Change category in the heartbeat to be a value instead of pointer.
* The undefined/unknown value is set to zero.
* The final marshalling will ignore if category is empty by omiting empty value.
* Fixed several tests where the category is `writing tests` because the entity has been detected as test files (__test.go_, _testdata_)

The category field behaves:
* If category is informed via flags then it uses it, otherwise defaults to undefined (0 value)
* If category is not informed via flags, it will try to detect category, if found uses it, otherwises remains undefined (0 value)
* If category comes from extra heartbeat it will try to parse a string and it can be any of the ones already defined
  * If the string is either string empty or "null", it will be undefined (0 value)
  * Otherwise it will use the category parsed

ps:. The most important layer is the category detection where removes `"coding"` from category if found.


pkg/params/params.go
```diff
@@ -177,24 +177,24 @@ type (
...
-		LineAdditions     any                 `json:"line_additions"`
-		LineDeletions     any                 `json:"line_deletions"`
-		LineNumber        any                 `json:"lineno"`
-		Lines             any                 `json:"lines"`
-		Project           string              `json:"project"`
-		ProjectAlternate  string              `json:"alternate_project"`
-		Time              any                 `json:"time"`
-		Timestamp         any                 `json:"timestamp"`
+		BranchAlternate   string  `json:"alternate_branch"`
+		Category          string  `json:"category"`
```
alanhamlett, line 181: We should omit this key in the API request payload when it's `coding` category, or when none detected or specified. That way the payload is smaller, and the server sets it to `Coding` category by default anyways.

alanhamlett (approved):

pkg/heartbeat/category.go
```diff
@@ -193,7 +201,8 @@ func (c *Category) UnmarshalJSON(v []byte) error {
 
 // MarshalJSON implements json.Marshaler interface.
 func (c Category) MarshalJSON() ([]byte, error) {
-	if c == UndefinedCategory {
+	// if the category is Undefined or Coding, we don't want to send it to the API
+	if c == UndefinedCategory || c == CodingCategory {
```
gandarez: @alanhamlett this solves the problem

cmd/offlineprint/testdata/offline_heartbeat.json
```diff
@@ -1 +1 @@
-[{"branch":"heartbeat","category":"coding","cursorpos":12,"dependencies":["dep1","dep2"],"entity":"/tmp/main.go","type":"file","is_write":true,"language":"Go","lineno":42,"lines":100,"project":"wakatime-cli","time":1592868367.219124,"user_agent":"wakatime/13.0.6"},{"branch":"summary","category":"debugging","cursorpos":13,"dependencies":["dep3","dep4"],"entity":"/tmp/main.py","type":"file","is_write":false,"language":"Python","lineno":43,"lines":101,"project":"wakatime","time":1592868386.079084,"user_agent":"wakatime/13.0.7"}]
\ No newline at end of file
+[{"branch":"heartbeat","category":"","cursorpos":12,"dependencies":["dep1","dep2"],"entity":"/tmp/main.go","type":"file","is_write":true,"language":"Go","lineno":42,"lines":100,"project":"wakatime-cli","time":1592868367.219124,"user_agent":"wakatime/13.0.6"},{"branch":"summary","category":"debugging","cursorpos":13,"dependencies":["dep3","dep4"],"entity":"/tmp/main.py","type":"file","is_write":false,"language":"Python","lineno":43,"lines":101,"project":"wakatime","time":1592868386.079084,"user_agent":"wakatime/13.0.7"}]
```
alanhamlett: We should omit the `"category"` key instead of sending empty string.

alanhamlett (approved):

PR #1206, opened 2025-07-25
Upgrade wsl to wsl_v5

hezhizhen:
```
WARN The linter 'wsl' is deprecated (since v2.2.0) due to: new major version. Replaced by wsl_v5.
WARN [linter] `allow-multiline-assign` is deprecated and always allowed in wsl >= v5
WARN Suggested new configuration:
linters:
  enable:
    - wsl_v5
  settings:
    wsl_v5:
      allow-first-in-block: true
      allow-whole-block: false
      branch-max-lines: 2
```

I updated `wsl` to `wsl_v5` and added the suggested configuration.

Some new lines were added automatically by wsl_v5.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,7 +19,7 @@ linters:
     - usestdlibvars
     - usetesting
     - whitespace
-    - wsl
+    - wsl_v5
   settings:
     revive:
       rules:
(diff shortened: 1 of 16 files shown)


alanhamlett (approved):
Thanks!

PR #1208, opened 2025-07-25
Fix integration tests

gandarez:
This PR fixes integration tests for extra heartbeats and add a new one `TestSendHeartbeats_SyncOfflineActivity`. It also moves the logic for sync max to params.


main_test.go
```diff
@@ -672,7 +671,111 @@ func TestSendHeartbeats_ExtraHeartbeats_SyncLegacyOfflineActivity(t *testing.T)
 
 	assert.Zero(t, offlineCount)
 
-	assert.Eventually(t, func() bool { return numCalls == 4 }, time.Second, 50*time.Millisecond)
+	assert.Eventually(t, func() bool { return numCalls == 3 }, time.Second, 50*time.Millisecond)
```
alanhamlett, line 740: We shouldn't increase the number of API requests, this should be at max 2:

```
1 heartbeat + 26 extra = 27 (0 requests)
- 25 sent to API = 2 saved to offline db (1 request)
- 2 offline heartbeats sent to API = 0 (2 requests)
```

gandarez: This is the scenario where we sync offline heartbeats from legacy file. Do you like to join the offline heartbeats with new and legacy files and call the API?

alanhamlett: No, it's fine as-is. Sorry, I thought this was increasing from 2 to 3 not decreasing from 4 to 3.

testdata/api_heartbeats_response_offline.json
```diff
@@ -0,0 +1,74 @@
+{
+    "responses": [
+        [
+            {
+                "data": {
+                    "branch": "heartbeat",
```
alanhamlett, line 6: The actual API only responds with the heartbeat's ID, no other attributes.

gandarez: All tests are expecting a heartbeat in the response. What's the payload returned?

alanhamlett:
```json
{
 "responses": [
           [
               {
                   "data": {
                       "id": "845a922e-9e65-4775-bd68-bb3196d2e06a"
                   }
               },
               201
          ], ...
  ]
}
```

alanhamlett: I've updated the mocked responses to match the API, but only for the tests we're changing here not all tests.

alanhamlett (approved):

testdata/api_heartbeats_response_offline.json
gandarez: @alanhamlett Im afk but I guess we don't parse the response correctly so we need to consider fixing it later in another PR.

alanhamlett: We must be parsing it correctly, or else the cli wouldn't be working in the wild. I changed it to mock only the id in responses here and it works, so looks good.

PR #1241, opened 2025-11-20
Add compact output option for today summaries

ZerGo0:
This PR adds a new compact output option that currently only affects the `--today` status bar output (I personally only care about this, but this could be used for other things in the future). This option can be set via the `--compact` flag or the `compact` setting in the `.wakatime` file. When enabled, today summaries return only the total time in text and JSON outputs instead of category breakdowns, reducing noisy status bar displays.

Includes documentation in USAGE and tests covering parameter loading precedence and summary rendering in both text and JSON modes. No behavior change occurs unless compact is explicitly enabled.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -32,6 +32,7 @@ include_only_with_project_file = false
 exclude_unknown_project = false
 status_bar_enabled = true
 status_bar_coding_activity = true
+compact = false
 status_bar_hide_categories = false
 status_bar_max_categories = 0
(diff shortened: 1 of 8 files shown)


gandarez:
In this case it has the same behavior as `hideCategories` flag.

ZerGo0:
Oh sorry, I'm blind apparently.

I'm going to close the PR, thanks for the quick response.

PR #1242, opened 2025-12-02
Prevent many random project names on project file write error

alanhamlett:
When we encounter an error saving the random project name to the project's `.wakatime-project` file, use `Unknown Project` or else we'll end up with a ton of random project names... one project name for each heartbeat.


alanhamlett:
Implemented with #1243 and #1244.

PR #1243, opened 2025-12-03
Prevent many random project names on project file write error

gandarez:
When we encounter an error saving the random project name to the project's `.wakatime-project` file, use `Unknown Project` or else we'll end up with a ton of random project names... one project name for each heartbeat. It also creates a new package `setup` under `log` to move the logic to setup logger from `cmd` and avoid import cycle not allowed.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -348,7 +348,11 @@ func obfuscateProjectName(ctx context.Context, folder string) string {
 
 	err := Write(folder, project)
 	if err != nil {
-		logger.Warnf("failed to write: %s", err)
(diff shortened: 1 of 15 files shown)


pkg/project/project.go
```diff
@@ -348,7 +348,11 @@ func obfuscateProjectName(ctx context.Context, folder string) string {
 
 	err := Write(folder, project)
 	if err != nil {
-		logger.Warnf("failed to write: %s", err)
+		logger.Errorf("failed to write: %s", err)
```
alanhamlett, line 351: 👍

alanhamlett (approved):
Awesome!

PR #1245, opened 2025-12-03
Add new languages

gandarez:
This PR adds new languages and enable a new linter to check missing keys on switch statements. It also fixes missing languages in tests.

- Core
- Gemtext
- Lox


alanhamlett (approved):

pkg/heartbeat/language.go
```diff
@@ -5780,16 +5801,17 @@ func (l Language) String() string {
...
 	default:
 		return languageUnknownStr
 	}
 }
 
 // StringChroma returns the corresponding chroma lexer name.
-// nolint:gocyclo
+// nolint:gocyclo,exhaustive
 func (l Language) StringChroma() string {
-	switch l {
```
alanhamlett, line 5792: why do we want to skip linting here?

alanhamlett: For `UnknownDetector` case?

gandarez: This function checks only few cases and adding all of them make it unfeasible. The default case is String().

PR #1253, opened 2026-01-03
feat (heartbeat): Class MDX files as documentation

GeekCornerGH:
MDX files are Markdown files supporting components. In my case, I use them with Docusaurus, but since it's Markdown, I believe the heartbeats should be in the "writing docs" category

### DIFF
--- a/pkg/heartbeat/category.go
+++ b/pkg/heartbeat/category.go
@@ -129,7 +129,7 @@ func DetectCategory(h Heartbeat) Category {
 		return WritingTestsCategory
 	}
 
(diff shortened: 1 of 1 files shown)


gandarez:
You must follow our guidelines otherwise CI fails

https://github.com/wakatime/wakatime-cli/blob/develop/CONTRIBUTING.md

GeekCornerGH:
Thank you @gandarez I just realized I should properly implement MDX instead of just classing files too. I'll be back with a proper PR!

PR #1254, opened 2026-01-05
Replace long switch stmt to language map

gandarez:
_AI was used to generate this summary_

**Changes Made**
1. Created `language_maps.go` (861 lines)
    * Contains stringToLanguage map: Maps normalized strings to Language constants (806 entries including aliases)
    * Contains chromaToLanguage map: Maps normalized chroma lexer names to Language constants (43 entries)
    * These are the source of truth for parsing
2. Created `language_generated.go` (853 lines)
    * Contains languageToString map: Maps Language constants to their canonical string (796 entries)
    * Contains languageToChroma map: Maps Language constants to chroma lexer names (41 entries)
    * Generated via go generate
3. Created `language_gen.go` (135 lines)
    * Generator script that reads language_maps.go and produces reverse maps
    * Run with go generate in the heartbeat package
4. Modified `language.go` (2546 lines, down from 5913)
    * Added //go:generate go run language_gen.go directive
    * Replaced ParseLanguage() switch (1600+ cases) with map lookup: O(1)
    * Replaced ParseLanguageFromChroma() switch (43 cases) with map lookup: O(1)
    * Replaced String() switch (796 cases) with map lookup: O(1)
    * Replaced StringChroma() switch (41 cases) with map lookup: O(1)

**Performance Improvement**
* **Before**: O(n) switch statements for all four functions
* **After**: O(1) map lookups

**File Size Reduction**
* language.go: 5913 → 2546 lines (**57% reduction**)
* Total package: Better organized with clear separation of concerns

**Benchmark**

Benchmark | Current | New | Change
-- | -- | -- | --
ParseLanguage | 294,086 ns/op | 504 ns/op | 🟢 583x faster
ParseLanguage_Single/common_language | 31,319 ns/op | 87.10 ns/op | 🟢 360x faster
ParseLanguage_Single/case_insensitive | 47,732 ns/op | 76.09 ns/op | 🟢 627x faster
ParseLanguage_Single/with_spaces | 72,759 ns/op | 111.3 ns/op | 🟢 654x faster
ParseLanguage_Single/unknown | 73,062 ns/op | 102.8 ns/op | 🟢 711x faster
StringChroma | 23.31 ns/op | 43.45 ns/op | 🔴 1.9x slower
StringChroma_Single/with_chroma_mapping | 1.884 ns/op | 5.040 ns/op | 🔴 2.7x slower
StringChroma_Single/without_chroma_mapping | 2.599 ns/op | 9.438 ns/op | 🔴 3.6x slower
StringChroma_Single/fallback_to_string | 2.565 ns/op | 9.496 ns/op | 🔴 3.7x slower

**Summary**

**Improvements (New Implementation)**
**ParseLanguage is dramatically faster** — The new implementation achieves **360x to 711x** speedup across all ParseLanguage benchmarks. This suggests a shift from a linear search or expensive lookup (likely iterating through Chroma's language registry) to a more efficient data structure like a pre-computed map.
* Microseconds → nanoseconds is a massive win for hot paths
* The improvement is consistent across all cases (common, case-insensitive, with spaces, unknown)

**Trade-offs (New Implementation)**
**StringChroma is slower** — The reverse lookup operation is **2-4x slower** in the new implementation:
* with_chroma_mapping: 1.9 ns → 5.0 ns
* without_chroma_mapping: 2.6 ns → 9.4 ns
* fallback_to_string: 2.6 ns → 9.5 ns
This is likely due to additional indirection or a different data structure that optimizes for parsing at the cost of serialization.

**Verdict**

**The new implementation is a clear net win** if ParseLanguage is called more frequently than StringChroma, which is typically the case (parsing happens on every heartbeat input, serialization happens less often or benefits from caching). Even in worst case scenarios:
* **ParseLanguage saved**: ~294,000 ns → ~500 ns = **~293,500 ns saved**
* **StringChroma cost**: ~23 ns → ~43 ns = **~20 ns added**
You'd need to call StringChroma **~14,000 times** for every ParseLanguage call to break even. The trade-off is negligible in practice.


pkg/heartbeat/language.go
```diff
@@ -2470,3433 +2471,39 @@ const (
 
 // ParseLanguage parses a language from a string. Will return false
 // as second parameter, if language could not be parsed.
-// nolint:gocyclo
 func ParseLanguage(s string) (Language, bool) {
-	switch normalizeString(s) {
```
alanhamlett, line 2475: I love how we don't have to manually maintain this list anymore!

alanhamlett (approved):

PR #1256, opened 2026-01-06
Generate dynamic language mapping

gandarez:
**Summary**

* Bump Chroma to latest version (v2.21.1).
* Changed `language_gen.go` to automatically generate test entries in `language_test.go`. The generator now:
  * **Reads all Chroma lexers and generates**:
    * `stringToLanguage` and `languageToString` maps (excluding "unknown"/`LanguageUnknown`)
    * `chromaToLanguage` and `languageToChroma` maps
  * **Generates `languageTests()` entries**: These are the primary string values for each Language constant (e.g., "Go" → `LanguageGo`). The entries respect priority overrides from the priority section in language.go.
  * **Generates `languageTestsAliases()` entries**: These are aliases that parse to Language constants but differ from the primary string. They include:
    * Chroma lexer names and aliases that differ from the primary Language string
    * Priority overrides (secondary strings that also map to the same Language)
    * chromaPriority entries
* **Added new Language constants**: C3, Kakoune, and RGBDS Assembly were new Chroma languages that needed to be added.
* **Added chromaPriority entries**: Protocol Buffer Text Format → LanguageProtocolBuffer, WebAssembly Text Format → LanguageWebAssembly
* **Updated test skip list**: Added entries to TestLanguage_StringChroma_AllLexersSupported for Chroma lexers that map to existing Language constants via chromaPriority (since StringChroma() returns the primary Chroma name, not the alias).

The generated test file now has:
**802 languageTests entries** (primary string values)
**212 languageTestsAliases entries** (aliases from Chroma and priority overrides)


alanhamlett (approved):
This is perfect!

PR #1261, opened 2026-01-12
Enable reading project placeholder from .wakatime-project file

gandarez:
This PR enables reading project placeholder from .wakatime-project file

It commit adds support for a `{project}` placeholder in `.wakatime-project` files, allowing users to dynamically include the folder name in their project configuration. 

Previously, users had to hardcode the full project name in their `.wakatime-project` file. This was limiting because:

* Organization prefixes - Many teams want to prefix project names with their company or team name (e.g., `my-company/wakatime-cli`) but didn't want to manually type the folder name
* Consistency - If you rename a folder, you'd have to manually update the `.wakatime-project` file to match
* Reusability - The same `.wakatime-project` template couldn't be reused across different projects

**How It Works Now**

With the `{project}` placeholder, users can dynamically include the folder name in their project configuration:

.wakatime-project content | Folder name | Result sent in heartbeat
--- | --- | --- |
my-company/{project} | wakatime-cli | my-company/wakatime-cli
{project}-backend | api | api-backend
{project} | myapp | myapp
team/{project}/main | dashboard | team/dashboard/main

This allows users to add **prefixes** (like organization names), **suffixes** (like environment indicators), or both, while still automatically picking up the actual folder name. The heartbeat sent to WakaTime will contain the fully resolved project name, making dashboards and reports more organized and consistent across teams.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -175,11 +175,25 @@ See the [source code](https://github.com/wakatime/wakatime-cli/blob/36f6372880d7
 
 ### WakaTime Project File
 
-To overwrite the auto-detected project, create a `.wakatime-project` file in your project’s root folder.
+To overwrite the auto-detected project, create a `.wakatime-project` file in your project's root folder.
 The first line of the file contents overwrites the project name, if present.
 The second line, if present, overwrites the current branch name when working inside this folder.
-When the `.wakatime-project` file is empty, the folder’s name is used as the project name.
+When the `.wakatime-project` file is empty, the folder's name is used as the project name.
 Whenever a `.wakatime-project` file is found, it overwrites all other project detection.
+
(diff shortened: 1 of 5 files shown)


alanhamlett:
Great feature!

alanhamlett:
How about using an empty file instead?

gandarez:
> How about using an empty file instead?

How would it work? Can you give some examples?

alanhamlett:
Ok, this is different. We [already use the folder as the project name](https://github.com/wakatime/wakatime-cli/blob/develop/USAGE.md#wakatime-project-file) when the file is empty but this is for adding prefixes or suffixes to the folder.

alanhamlett (approved):

PR #1265, opened 2026-01-15
feat (heartbeat): Improve test files detection

GeekCornerGH:
**NOT YET TESTED**, I made this PR on the go (pun not intended). I'll of course test it later today

Reviews are still welcome tho

PR #1266, opened 2026-01-15
feat (heartbeat): Improve test files detection

GeekCornerGH:
#1265, but with a proper branch name

**NOT YET TESTED**, I made this PR on the go (pun not intended). I'll of course test it later today

Reviews are still welcome tho

### DIFF
--- a/pkg/heartbeat/category.go
+++ b/pkg/heartbeat/category.go
@@ -3,6 +3,7 @@ package heartbeat
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


gandarez:
@alanhamlett wdyt?

alanhamlett:
I want to make sure it doesn't match non-test file paths. Looks ok, right?

GeekCornerGH:
@alanhamlett sorry for the delay. I tested it today, and it worked fine in my case. I've also tested the regex against https://regex101.com and wasn't able to spot any false positives.

alanhamlett:
What about changing

```
regexp.MustCompile(`(?i).*\.(test|spec)\.[^.]+$`)
```

to:

```
regexp.MustCompile(`(?i).*[\._](test|spec)\.[^./\\]+$`)
```

GeekCornerGH:
Good call, let me do it

alanhamlett (approved):

alanhamlett:
Looks great!

gandarez:
Can you create another PR to move the regex to a global variable? There are other examples in the project.

PR #1268, opened 2026-01-20
Add support for multiple api urls

### LINKED ISSUE #1267: Allow mutli account tracking
I found out about Wakatime from Hackclub who use their own fork - Hackatime, and subsequently since then I've been using their API key in my config. However, now I want to track my time on both Hackatime and Wakatime, so sending heartbeats to both. 

According to https://github.com/wakatime/jetbrains-wakatime/issues/156 there's currently no way to do it directly, has that changed since then? If not, could you consider options in the config that allow for that as a feature? I do think quite a few people might need it.

And as discussed in the issue I mentioned before, as an alternative I could try syncing their heartbeats with the api, which is what I'll do otherwise.

gandarez:
This commit implements the `[api_urls]` configuration section, allowing users to route heartbeats to different WakaTime-compatible API servers based on file path patterns. This is useful for sending work-related coding activity to a corporate server while personal projects go to the default WakaTime API.

**Config Format**

`[api_urls]`
`^/home/user/work/ = https://work.example.com/api/v1|your-work-api-key`
`^/home/user/personal/ = https://personal.example.com/api/v1|your-personal-api-key`

The format is `pattern = api_url|api_key`, where:

**pattern**: A regex matched against file paths (case-insensitive)
**api_url**: The WakaTime API base URL
**api_key**: The API key for that server

**Sumamry**
1. URL Pattern Matching: Patterns are compiled as case-insensitive regex and matched against file paths
2. Heartbeat Grouping: Heartbeats are grouped by `APIURL|APIKey` combination before sending
3. Priority: `project_api_key` is checked first to override the api key for the default destination, then `api_urls` patterns add additional destinations
4. URL Normalization: API URLs are normalized (trailing `/users/current/heartbeats.bulk` is stripped)
5. Validation: API keys are validated using the same rules as the main `api_key` setting

closes https://github.com/wakatime/wakatime-cli/issues/1267


pkg/params/params.go
```diff
@@ -298,38 +354,98 @@ func LoadAPIParams(ctx context.Context, v *viper.Viper, order FlagReadOrder) (AP
...
+		}
+
+		compiled, err := regex.Compile(k)
+		if err != nil {
+			logger.Warnf("failed to compile api_urls regex pattern %q", k)
+			continue
+		}
+
+		// split value by | to get api_url and api_key
+		parts := strings.SplitN(s, "|", 2)
```
alanhamlett, line 393: Default to wakatime api url when only `len(parts) == 1`. For ex:

`.* = your-personal-api-key`

pkg/params/params.go
```diff
@@ -279,9 +280,64 @@ func LoadAPIParams(ctx context.Context, v *viper.Viper, order FlagReadOrder) (AP
...
+		}
+	}
+
+	timeout := api.DefaultTimeoutSecs
+
+	if timeoutSecs, ok := vipertools.FirstNonEmptyInt(v, "timeout", "settings.timeout"); ok {
+		timeout = timeoutSecs
+	}
+
+	return API{
```
alanhamlett, line 321: Let's combine all these into one `URLPatterns`:

* `Key`
* `KeyPatterns`
* `URL`
* `URLPatterns`

That will be type `[]apikey.URLPattern`. We build `URLPatterns` by first `make(map[string]apikey.URLPattern)` then appending each map value to the patterns array at the end. The map keys are `api_key:api_url`. Here's the process:

1. read `api_url` and `api_key`. Default `api_url` to wakatime if not defined. Add to map.
2. read `api_keys` and using `api_url` from step 1, add any api keys to the map.
3. read `api_urls` and add all to the map, using `api_url` from step 1 if the line only has a pattern and key
4. now turn the map's values into the array and assign to `API{URLPatterns}` for returning

alanhamlett (approved):

PR #1273, opened 2026-02-25
Switch to gandarez semver-action

alanhamlett:


gandarez:
Let's wait to merge and do all necessary's tests to make sure it works out of the box

Issue #1288, opened 2026-03-29
Excessive CPU Usage in v2 / Repeatedly failing to parse Claude transcripts


JoeyEamigh:
Since the v2 upgrade, `wakatime` has been using ~8% CPU on my 9950X. There is an instance of `wakatime` running every time I open or edit a file, and it seems it will attempt to parse `claude` transcripts repeatedly, but fail to do so. With 2 `claude` instances and `vscode` running, this spawns 3 `wakatime` instances, each trying to parse this data.

**Environment**:

- OS: Arch Linux
- Platform: x86_64

**Logs**:

This is spammed in the log a few thousand times:

```
{"level":"warn","now":"2026-03-29T01:18:51-04:00","caller":"ai/claude.go:176","func":"github.com/wakatime/wakatime-cli/pkg/ai.Claude.parseTranscript","message":"failed parsing claude transcript line from: "[claude file]": json: cannot unmarshal string into Go struct field claudeLogLine.toolUseResult of type ai.toolUseResult","version":"v2.0.0","os/arch":"linux/amd64","file":"[editor file]","time":1774761416.714,"plugin":"vscode/1.113.0 vscode-wakatime/30.0.0","lineno":5,"is_write":true}
```

mochafreddo:
I think PR #1289 fixes one of the Claude transcript parsing failures, but not all of them.

It should handle:
- json: cannot unmarshal string into Go struct field claudeLogLine.toolUseResult of type ai.toolUseResult

However I’m still seeing parse failures for a different shape after that fix:

- json: cannot unmarshal array into Go struct field toolUseResult.toolUseResult.content of type string

My environment:
- macOS arm64
- Cursor 1.105.1
- vscode-wakatime 29.0.3
- wakatime-cli v2.0.0

This still causes repeated warnings and keeps the extension host hot because the same Claude transcript files are parsed over and over.

It looks like `pkg/ai/claude.go` still assumes `toolUseResult.content` is always a string, but some Claude transcript lines now emit `content` as an array.

I can put together a follow-up PR if helpful. My current thought is to make `content` tolerant to both string and array shapes, similar to what #1289 did for `toolUseResult`.

alanhamlett:
Let's check if [v2.0.5](https://github.com/wakatime/wakatime-cli/releases/tag/v2.0.5) fixes it. To force an update, run this command then restart your IDE:

```
rm ~/.wakatime/wakatime-cli*
```

[...]

PR #1292, opened 2026-03-29
[codex] handle Claude content arrays

mochafreddo:
This follows up on the Claude transcript parser failures discussed in #1288.

PR #1289 fixed one important case where `toolUseResult` itself could be a string. However, Claude transcripts can still contain summary-style `toolUseResult` payloads where `content` is an array instead of a string. In that case the current parser fails during JSON unmarshalling before it has a chance to inspect the rest of the record, which leads to repeated warning spam and extra CPU churn from re-parsing the same transcript files.

This change makes `toolUseResult.content` and `toolUseResult.file.content` tolerant to either a string or an array payload. The parser still only computes line deltas from string content, so behavior for actual file-edit heartbeats stays the same. Array-shaped content is simply accepted and ignored when it does not map to an editable file heartbeat.

I also added a Claude transcript test case covering a `toolUseResult.content` array with text blocks, matching the shape I reproduced locally from Cursor/Claude transcripts. The targeted validation used was:

`mise exec go@1.25.5 -- go test ./pkg/ai/...`

### DIFF
--- a/pkg/ai/claude.go
+++ b/pkg/ai/claude.go
@@ -25,15 +25,20 @@ type (
 		OldLines int `json:"oldLines"`
 	}
 
+	contentValue struct {
+		String *string
+		Array  *[]json.RawMessage
+	}
+
 	toolUseResultFile struct {
-		Content  *string `json:"content"`
-		FilePath *string `json:"filePath"`
+		Content  *contentValue `json:"content"`
+		FilePath *string       `json:"filePath"`
 	}
 
 	toolUseResult struct {
 		Type            *string            `json:"type"`
 		File            *toolUseResultFile `json:"file"`
-		Content         *string            `json:"content"`
+		Content         *contentValue      `json:"content"`
 		FilePath        *string            `json:"filePath"`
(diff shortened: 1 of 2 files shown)


mochafreddo:
Follow-up from a local repro on Cursor + vscode-wakatime.

PR #1289 fixed the case where `toolUseResult` itself can be a string, but I was still hitting parse failures when `toolUseResult.content` came through as an array of text blocks.

This patch makes Claude transcript parsing tolerate both string and array shapes for `content`, so those lines no longer fail JSON unmarshalling and spam warnings. For file-edit heartbeats, it also counts line changes from array text blocks instead of dropping them to zero.

Added targeted test coverage for:
- summary-style `toolUseResult.content` arrays
- file-edit `toolUseResult.content` arrays

Validation used:
`mise exec go@1.25.5 -- go test ./pkg/ai/...`

alanhamlett:
We require branch names prefixed with `major/`, `minor/`, `bugfix/`, or `patch/` for our automatic versioning logic. I've re-created this PR in #1295.

PR #1295, opened 2026-03-29
Handle Claude content arrays

alanhamlett:
Same as #1292 but with required branch name.


mochafreddo:
Thanks for handling the branch naming mismatch and re-creating this so quickly. I appreciate it.

Issue #1423, opened 2026-05-19
High cpu and high disk r/w


bookshiyi:
Every few minutes, wakatime will take up a large amount of CPU and high-speed hard disk r/w.
Is this really a bug? Or malicious behavior

**Environment**:

- OS: macOS
- Platform: arm64

**Logs**:
Process name "5.28.3"

<img width="266" height="175" alt="Image" src="https://github.com/user-attachments/assets/c87438ff-ad5c-411b-afc8-4c89f79e7fe4" />
<img width="517" height="83" alt="Image" src="https://github.com/user-attachments/assets/e8846c96-724f-4c82-a8b5-a92c1c558f75" />

alanhamlett:
My guess is when we parse Cursor sqlite logs looking for AI codegen, we process too many sql rows. I've improved that by limiting the rows in the Cursor sql query with #1424.

PR #1424, opened 2026-05-19
Improve Cursor sqlite query performance

alanhamlett:
To reduce CPU usage for #1423.


bookshiyi:
not only cursor, opencode also

PR #1508, opened 2026-07-14
fix: prevent permanent backoff by capping retries and resetting stale state

### LINKED ISSUE #1283: Exponential backoff counter never auto-resets when server becomes reachable again
### Environment

- **OS:** Windows 11 (10.0.22621)
- **Platform:** amd64
- **wakatime-cli version:** v1.139.4
- **API server:** Wakapi (self-hosted at wakapi.dev)
- **Integration:** claude-code-wakatime v3.0.5 (Claude Code hook)

### Description

The exponential backoff mechanism accumulates retries indefinitely and never auto-resets when the API server becomes reachable again. Once the counter reaches a high value, the CLI effectively stops sending heartbeats permanently until the user manually resets `~/.wakatime/wakatime-internal.cfg`.

### Steps to Reproduce

1. Configure wakatime-cli with an API server (Wakapi or WakaTime cloud)
2. Let the server become temporarily unreachable (PC sleep, server restart, brief network outage)
3. CLI enters exponential backoff (expected)
4. Server comes back online
5. CLI remains in backoff — never attempts a fresh connection

### What Happened

After a brief server outage, `wakatime-internal.cfg` showed:

```ini
[internal]
backoff_at      = 2026-03-23T09:38:19+05:30
backoff_retries = 2839
```

With 2,839 retries, the backoff window is so large that every heartbeat call returns exit code 1 immediately without attempting to reach the API. The CLI is effectively dead.

When used as a Claude Code hook, this produces `PreToolUse hook error` on every tool call — hundreds of errors per session.

### What I Expected

The backoff counter should auto-reset after:
- A maximum retry count (e.g., cap at 10, then try fresh)
- A maximum backoff duration (e.g., max 1 hour between attempts)
- A time-based reset (e.g., if `backoff_at` is older than 24 hours, reset)
- Or: attempt a real health check every N minutes regardless of backoff state, reset on success

### Workaround

Manually resetting the internal config fixes it immediately:

```bash
sed -i 's/backoff_retries.*=.*/backoff_retries           = 0/' ~/.wakatime/wakatime-internal.cfg
sed -i 's/backoff_at.*=.*/backoff_at                = /' ~/.wakatime/wakatime-internal.cfg
```

### Logs

Debug output from `wakatime-cli --log-to-stdout --verbose` after the backoff accumulated:

```
{"level":"debug","message":"params: api params: (backoff at: '2026-03-23T09:38:19+05:30', backoff retries: 2839, ...)"}
```

The CLI checks the backoff state, sees it's within the backoff window, and exits without trying.

(Note: the troubleshooting docs say to delete `wakatime-internal.cfg` and filter out backoff errors — but in this case the backoff itself IS the bug, not a symptom of another error.)

### Related Issues

- #785 (Exit code 112, working offline)
- #535 (Separate file for internal config)
- wakatime/vscode-wakatime#257 ("working offline" stuck state)
- wakatime/vscode-wakatime#430 ("won't send heartbeat due to backoff")

Jonathanthedeveloper:
## Summary

Three fixes to prevent WakaTime from permanently stopping sync after temporary network outages or API failures.

**Fixes #1283**

### 1. Cap backoff retries at 10 (`pkg/backoff/backoff.go`)
- Added `maxRetries = 10` constant
- `WithBackoff` now caps retries at 10 when incrementing, preventing unbounded growth of the retry counter (which previously grew to 2839+ in issue #1283)

### 2. Add 24-hour safety reset in `shouldBackoff()` (`pkg/backoff/backoff.go`)
- If `backoff_at` is older than 24 hours, force-return `false` to reset backoff state
- This handles edge cases where backoff state was never properly cleared (e.g., system clock jumps, corrupted state)
- Note: The existing `parsed.After(time.Now())` check in `loadBackoffParams` already handles future dates; this adds a complementary check for stale past dates

### 3. Always run offline sync (`cmd/run.go`)
- `RunCmdWithOfflineSync` now always runs `offlinesync.RunWithRateLimiting`, even when the heartbeat command fails
- Previously, if heartbeat failed (e.g., due to backoff), offline sync was skipped entirely, so queued heartbeats could never drain even when the API became reachable

## Root Cause

The bug was a combination of:
1. `shouldBackoff()` logging "will reset" but never actually resetting backoff state when `backoffSeconds > maxBackoffSecs`
2. No cap on retry counter growth
3. Offline sync being skipped when heartbeat fails, creating a deadlock where heartbeats accumulate but never drain

## Testing

- Added 2 tests for the 24-hour stale backoff reset (`TestShouldBackoff_StaleBackoffResetsAfter24Hours`, `TestShouldBackoff_StaleBackoffActiveWithin24Hours`)
- All existing tests pass
- `go build ./...` succeeds

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -239,14 +239,21 @@ func RunCmd(ctx context.Context, v *viper.Viper, verbose bool, sendDiagsOnErrors
 }
 
 // RunCmdWithOfflineSync runs a command function and exits with the exit code
-// returned by the command function. If command run was successful, it will execute
-// offline sync command afterwards. Will send diagnostic on any errors or panics.
+// returned by the command function. It will always execute offline sync afterwards,
+// regardless of whether the command succeeded or failed. Will send diagnostic on
+// any errors or panics.
 func RunCmdWithOfflineSync(ctx context.Context, v *viper.Viper, verbose bool, sendDiagsOnErrors bool, cmd cmdFn) error {
-	if err := runCmd(ctx, v, verbose, sendDiagsOnErrors, cmd); err != nil {
-		return err
+	heartbeatErr := runCmd(ctx, v, verbose, sendDiagsOnErrors, cmd)
+
+	// always attempt offline sync, even when heartbeat command fails,
+	// to ensure queued heartbeats are drained when the API becomes reachable
+	syncErr := runCmd(ctx, v, verbose, sendDiagsOnErrors, offlinesync.RunWithRateLimiting)
+
+	if heartbeatErr != nil {
+		return heartbeatErr
 	}
 
-	return runCmd(ctx, v, verbose, sendDiagsOnErrors, offlinesync.RunWithRateLimiting)
+	return syncErr
 }
 
 // runCmd contains the main logic of RunCmd.
--- a/pkg/backoff/backoff.go
+++ b/pkg/backoff/backoff.go
@@ -19,6 +19,8 @@ import (
 const (
 	// maxBackoff sets the maximum seconds we will rate limit before retrying to send.
 	maxBackoffSecs = 3600
+	// maxRetries sets the maximum number of retries before resetting backoff state.
+	maxRetries = 10
 	// factor is the total seconds to be multiplied by.
 	factor = 15
 )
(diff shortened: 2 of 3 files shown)


alanhamlett:
The first two aren't necessary because backoff only applies when less than 3600 seconds, and the third is just wrong.

PR #1515, opened 2026-07-22
Fix AI transcript project overrides

skyfallwastaken:
## Summary

Keep AI transcript project context when sync is triggered by an ordinary heartbeat.

Explicit `--sync-ai-activity --project` behavior is unchanged.

## Tests

- Added unit and end-to-end regression coverage
- Passed `make test-all`


alanhamlett:
Not needed, we use `--alternate-project` not `--project` in all plugins.

PR #1518, opened 2026-07-24
feat(framework): detect JS/TS framework from nearest package.json

i243137-oss:
## Summary

Adds framework detection alongside the existing language detection. Given a heartbeat
for a JS/TS-family file, it walks up from the file to the nearest `package.json` and
checks `dependencies`/`devDependencies` against a small registry of known frameworks,
setting a new `framework` field on the heartbeat.

Detected frameworks: React, Next.js, Vue, Nuxt, Angular, NestJS, Svelte, SvelteKit,
SolidJS, Preact, Qwik, Remix, Astro.

## Why a new field instead of changing `language`

`Heartbeat.Language` is sent as-is to the API and drives existing language statistics.
Overwriting `JSX` → `React` would silently corrupt users' language stats. `framework`
is additive (`omitempty`), so behavior for every existing heartbeat is unchanged.

## Design

- `pkg/framework` — a small `Framework{Name, Dependency}` registry; adding support for
  another framework is a one-line addition, no new conditionals.
- Runs after `language.WithDetection` in the pipeline (needs `Heartbeat.Language`
  already set) — wired via a new `WithFrameworkDetection()` preprocessor.
- Reuses `pkg/file.Find` for the upward `package.json` search (the same helper
  `language`/`project` already use).
- `package.json` reads/parses are memoized per run, so multiple heartbeats for files in
  the same project only pay the parse cost once.
- Registry is ordered meta-framework-first (e.g. Next.js before React, Nuxt before Vue,
  SvelteKit before Svelte) so a project depending on both resolves to the more specific
  one.
- Missing/malformed/unreadable `package.json` → no framework detected, no error, no
  panic. Existing behavior is fully preserved when nothing is detected.

## Not included

Backend/API support for storing `framework` — that lives in a separate repo. This PR
is CLI-side groundwork only; the field is inert until the API recognizes it, similar to
how other newer heartbeat fields appear to have been introduced CLI-first.

## Testing

Table-driven tests in `pkg/framework/framework_internal_test.go` covering:
- All 13 supported frameworks
- Plain JS/JSX with no framework dependency
- Missing `package.json`
- Invalid/malformed `package.json`
- Nested `package.json` lookup (finds nearest, not farthest)
- Detection via `devDependencies`
- Cache correctness (memoizes both successful and failed parses)
- Language gating (skips non-JS/TS languages)
- `LocalFile` vs `Entity` precedence
- Full `WithDetection()` pipeline integration, including backward compatibility when no
  framework is detected

`go test ./...` passes for every package this change touches
(`pkg/framework`, `pkg/heartbeat`, `cmd/handler`, `cmd/heartbeat`).

## Known unrelated issue

`TestSyncOfflineActivity_MultipleApiKey` in `cmd/offlinesync` fails locally on Windows
due to a `bbolt` file handle still being open when `t.TempDir()` cleanup tries to
delete it — reproduces on a clean checkout without this change and is unrelated to
`pkg/framework`.

### DIFF
--- a/cmd/handler/option.go
+++ b/cmd/handler/option.go
@@ -6,6 +6,7 @@ import (
 	"github.com/wakatime/wakatime-cli/pkg/fileexperts"
 	"github.com/wakatime/wakatime-cli/pkg/filestats"
 	"github.com/wakatime/wakatime-cli/pkg/filter"
+	"github.com/wakatime/wakatime-cli/pkg/framework"
 	"github.com/wakatime/wakatime-cli/pkg/heartbeat"
 	"github.com/wakatime/wakatime-cli/pkg/language"
 	"github.com/wakatime/wakatime-cli/pkg/params"
@@ -81,6 +82,13 @@ func WithLanguageDetection() Preprocessor {
 		return language.WithDetection(language.Config{
 			GuessLanguage: params.Heartbeat.GuessLanguage,
 		})
+	}
+}
+
+// WithFrameworkDetection returns a Preprocessor that applies framework detection to heartbeats.
+func WithFrameworkDetection() Preprocessor {
+	return func(_ params.Params) heartbeat.HandleOption {
+		return framework.WithDetection()
 	}
 }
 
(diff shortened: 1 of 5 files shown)


1chooo (commented):
This feature looks great! Should we loop in the frontend team, though? Since we might add more fields to the screen, we'll want to make sure everything stays consistent across the UI.

alanhamlett:
We already have this, it's called WithDependencyDetection so we don't need to duplicate it.

PR #1536, opened 2026-08-24
feat: add Codex source override

carlosflorencio:
**Describe the pull request**

This PR adds an opt-in `--codex-source` flag and `WAKATIME_CODEX_SOURCE` environment variable for AI heartbeat sync. When either is set, the Codex parser uses the supplied source for editor attribution instead of the source recorded in the transcript. The explicit flag takes precedence over the environment variable. If neither is set, the existing transcript-source behavior is unchanged.

When an application runs `codex-acp`, and `codex-acp` starts Codex app-server, the transcript source is currently hardcoded to `vscode`. WakaTime uses that source for editor attribution, so it reports the session as `Codex VSCode` even when another application, such as Kandev, is the tool actually invoking Codex.

The better long-term fix is for upstream `codex-acp` and Codex to support a dynamic source supplied by the application that starts them. This underlying Codex issue is already tracked in [openai/codex#23442](https://github.com/openai/codex/issues/23442). This WakaTime override is temporary support while we wait for a maintainer fix upstream. Once that support is available, integrations such as `codex-acp` can pass the correct source directly. Until then, this override lets the invoking application identify itself so WakaTime can report `codex-kandev/<version>` correctly without changing Codex transcripts.

For the current Kandev ACP workaround, configure the environment inherited by the WakaTime plugin and CLI:

```sh
WAKATIME_CODEX_SOURCE=kandev
```

The WakaTime Codex plugin already forwards its environment to `wakatime-cli`.

**Testing**

- `make test`
- `make test-integration`
- `make test-ip`
- `go test ./pkg/ai ./pkg/params ./cmd ./cmd/heartbeat -count=1`
- `golangci-lint run ./pkg/ai ./pkg/params ./cmd ./cmd/heartbeat`
- Shell tests were not run because `bats` is not installed in the environment.

The full `make test-all` target is also blocked by two existing `gosec` findings in `pkg/api/heartbeat.go`.

### DIFF
--- a/cmd/heartbeat/heartbeat.go
+++ b/cmd/heartbeat/heartbeat.go
@@ -228,6 +228,7 @@ func applyAIParsing(
 	heartbeats []heartbeat.Heartbeat,
 ) ([]heartbeat.Heartbeat, error) {
 	handle := ai.WithAISync(ai.Config{
+		CodexSource:  params.AI.CodexSource,
 		SyncDisabled: params.AI.SyncDisabled,
 		Plugin:       params.API.Plugin,
 		Project:      params.Heartbeat.Project,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -215,6 +215,7 @@ func setFlags(cmd *cobra.Command, v *viper.Viper) {
 		"",
 		"Format output. Can be \"text\", \"json\" or \"raw-json\". Defaults to \"text\".",
 	)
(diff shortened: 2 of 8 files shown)


alanhamlett:
We don't want short-term solutions in this repo. Do the better fix in the upstream plugin that calls wakatime-cli.

PR #1537, opened 2026-08-27
Respect CODEX_HOME env

ch4og:
Use CODEX_HOME for Codex sessions while preserving the default path when
it is unset.

Even when `CODEX_HOME` is set, the Codex parser still scans the hardcoded
`$HOME/.codex/sessions` path and misses sessions stored under
`$CODEX_HOME/sessions`. This PR makes the parser use `CODEX_HOME`
while keeping the default path when the environment variable is unset.


alanhamlett:
Closed with #1539

PR #1549, opened 2026-08-30
fix(ai): track Claude transcript cutoff per file

### LINKED ISSUE #1548: Claude Code AI sync drops ~65% of token usage with concurrent sessions (global ai_logs_last_parsed_at cutoff)
## Summary

`--sync-ai-activity` for Claude Code reports only a fraction of the token usage present in the transcripts when several Claude Code sessions or subagents run concurrently. Over a 7 day window WakaTime received ~1.07B cached input / 1.95M output tokens while the transcripts contain ~3.5B / 6.1M (Claude models only, cross-checked with ccusage). Cost estimates are proportionally low.

## Cause

`pkg/ai/ai.go` stores a single global cutoff `internal.ai_logs_last_parsed_at`, set to the newest heartbeat time across all parsed transcripts. In `pkg/ai/claude.go` `parseTranscript`, every line with a timestamp before that cutoff goes through `advanceTokens`, which discards its usage delta.

Two things are lost on every sync (every ~60s from the Claude Code plugin):

1. The tail of every transcript: usage logged after the transcript's last heartbeat (thinking, reads, Bash, prompts with no file entity yet). It has not been attached to any heartbeat, but on the next run it is before the cutoff and skipped.
2. Everything in slower parallel sessions and subagents whose lines predate the newest session's heartbeat.

## Reproduction

1. Replay a full day with the cutoff set to that day's midnight: totals match the transcripts (~91% of ccusage, remaining diff is timezone edges).
2. Let the plugin sync incrementally with several sessions open: server totals end up at roughly a third.

Minimal unit reproduction: a transcript with a heartbeat line at 12:00, a usage-only line at 12:30, then the global cutoff moved to 12:45 by another session, then a heartbeat line at 13:00. The 12:30 usage never reaches any heartbeat.

## Fix

Track the newest heartbeat time per transcript (`~/.wakatime/ai-claude-transcripts.json`) and resume each transcript right after it (exclusive), so pending usage rolls into the next heartbeat. Unseen transcripts keep using the global cutoff. Verified against a 2 day replay: identical totals to the current binary on a full scan, and a second run right after produces only new heartbeats. PR follows.

The same pattern likely affects the other transcript-based parsers, but I only verified and patched Claude.

samuelpatro:
Fixes #1548.

The global `ai_logs_last_parsed_at` cutoff is the newest heartbeat across all transcripts. With concurrent Claude Code sessions and subagents, lines of slower sessions that predate it are skipped on the next sync and their usage is discarded by `advanceTokens`. Usage logged after a transcript's last heartbeat is lost the same way on every sync. In practice ~65% of tokens never reached the server.

This persists a per-transcript checkpoint next to the internal config file (`<internal config>-ai-claude-transcripts.json`): the newest heartbeat timestamp, how many heartbeats were generated at that timestamp (so records appended later with an equal timestamp are still counted, without re-sending), and the transcript's size and mtime (so unchanged transcripts are skipped by a stat, without being opened). Each transcript resumes from its own checkpoint, so pending usage rolls into the next heartbeat. Unseen transcripts still use the global cutoff, so first runs and backfills behave as before. The state file is only rewritten when something changed, and entries for deleted transcripts are pruned.

Benchmarks: `--sync-ai-activity` against a local HTTP stub, 4.4 GB `~/.claude/projects` corpus (3,687 transcripts), Apple Silicon, isolated `WAKATIME_HOME`. Backfill is the first run after resetting the cutoff N days back; steady is the median of 5 immediate reruns, which is what the Claude Code plugin runs every 60 seconds.

| build | backfill 7d | steady 7d | backfill 30d | steady 30d |
|---|---|---|---|---|
| v2.26.0 upstream | 43.8 s | 0.47 s | 278 s | 0.47 s |
| this PR | 35.2 s | 0.29 s | 289 s | 0.30 s |

Steady state is faster than upstream because upstream re-parses every transcript modified after the global cutoff (all active sessions) on each run, while this only opens transcripts whose size or mtime changed. Backfill is unchanged within noise.

`BenchmarkClaudeParseSteadyState` (added; 200 transcripts x 300 lines, state populated): 0.63 ms and 0.3 MB allocated per parse, versus 349 ms and 206 MB for a checkpoint without the stat skip.

Correctness replay from a fixed cutoff with the offline queue drained: per-day token totals identical to upstream (e.g. 2026-08-30: 3,958,014 input / 105,681,576 cached / 422,819 output, 205 heartbeats), and an immediate second run sends 0 heartbeats, where upstream re-sends its last one each sync.

Other transcript-based parsers likely share the issue; scoped to Claude here since that is what I could verify.


alanhamlett:
• Found two correctness issues:

  - [P2] Namespace transcript state by --internal-config — claude.go L1414
    The global cutoff and synchronization lock are scoped to the configured internal file, but the new Claude state always uses ~/.wakatime/ai-claude-transcripts.json. Invocations using different internal configs therefore
    share Claude cutoffs and can concurrently overwrite the same .tmp file. Activity parsed under one profile can consequently be skipped under another. The state location or key should follow the same namespace as
    ini.InternalFilePath.

  - [P2] An exclusive timestamp watermark drops late equal-timestamp records — claude.go L1394
    Claude timestamps have finite precision. If the active writer appends a record whose timestamp equals the saved heartbeat timestamp after the previous scan reached EOF, the next scan processes its usage and immediately
    discards it because timestamp.After(cutoff) is false. A later heartbeat will never receive those tokens. The checkpoint needs a position/record-identity tie-breaker, or the boundary must remain inclusive with
    deduplication.

samuelpatro:
Both addressed in 33f3631 (merged develop in rather than rebasing to keep the discussion anchored):

- The state path now derives from `ini.InternalFilePath`: `<internal config without extension>-ai-claude-transcripts.json`, so separate `--internal-config` invocations get separate state, and writes go through `os.CreateTemp` + rename instead of a fixed `.tmp` name.
- The checkpoint boundary is inclusive again. Each checkpoint stores the newest heartbeat timestamp plus how many heartbeats were generated with exactly that timestamp; on the next scan, records at the checkpoint timestamp are re-parsed and only that many heartbeats are skipped. A record appended later with an equal timestamp keeps its heartbeat and its token usage rolls into the next one. Covered by the third pass in `TestClaudeParse_PerTranscriptCutoff`.

Patch coverage should also look better now: pkg/ai is at 89.9% locally, with the remaining uncovered lines being the temp-file write/close error branches.

samuelpatro:
Pushed ad79a71 with a performance fix for a regression I introduced in the first revision, plus benchmarks.

The per-transcript checkpoint on its own caused every tracked transcript to be re-read on every sync: a transcript's mtime is always at or after its checkpoint, since the last heartbeat line is usually also the last write, so the mtime filter never excluded it. The checkpoint now also records the transcript's size and mtime, and a transcript whose stat still matches is skipped without being opened. The state file is only rewritten when something changed, and stale entries are pruned from the paths seen during the walk rather than a stat per entry.

Measured with `--sync-ai-activity` against a local HTTP stub, on a 4.4 GB `~/.claude/projects` corpus (3,687 transcripts), Apple Silicon, isolated `WAKATIME_HOME`. "Backfill" is the first run after resetting `ai_logs_last_parsed_at` to N days back; "steady" is the median of 5 immediate reruns, which is what the Claude Code plugin does every 60 seconds.

| build | state | backfill 7d | steady 7d | backfill 30d | steady 30d |
|---|---|---|---|---|---|
| v2.26.0 upstream | none | 43.8 s | 0.47 s | 278 s | 0.47 s |
| 33f3631 (previous) | 161 / 772 transcripts | 36.5 s | 2.72 s | 306 s | 9.3 s |
| ad79a71 (this) | 188 / 937 transcripts | 35.2 s | 0.29 s | 289 s | 0.30 s |

Backfill cost is unchanged within noise. Steady-state is now faster than upstream because upstream re-parses every transcript modified after the global cutoff (all currently active sessions) on each run, while this only opens transcripts whose size or mtime changed; the remaining 0.3 s is the directory walk plus process startup.

`BenchmarkClaudeParseSteadyState` (added, 200 transcripts x 300 lines, state already populated): 349 ms and 206 MB allocated per parse at 33f3631, 0.63 ms and 0.3 MB at ad79a71.

Correctness replay from a fixed cutoff, offline queue drained: token totals per day identical to upstream (e.g. 2026-08-30: 3,958,014 input / 105,681,576 cached / 422,819 output, 205 heartbeats), and an immediate second run sends 0 heartbeats, where upstream re-sends its last one each sync.

golangci-lint v2.11 passes with 0 issues (run with the go 1.26 toolchain). pkg/ai coverage is 89.9%.

alanhamlett (changes requested):
• The two original findings are mostly addressed, but I found three remaining issues in head ad79a717:

  - [P1] Build expected paths with filepath.Join — claude_internal_test.go L25-L33
    These assertions append / manually, while production uses filepath.Join. On Windows, t.TempDir() and the result use backslashes, so at least the default-path assertion fails. Construct both config and expected paths
    with filepath.Join.

  - [P1] Do not use directory permission bits to force a Windows failure — claude_internal_test.go L104-L115
    Windows’ syscall.Mkdir ignores the supplied Unix mode, so the 0500 directory remains writable. saveTranscriptState creates state.json, making assert.NoFileExists fail. This error path needs a platform-independent setup
    or an OS-specific test.

  - [P2] Preserve the full internal-config filename in the namespace — claude.go L1483-L1489
    Removing the extension makes distinct configurations collide: work.cfg, work.ini, and extensionless work all map to work-ai-claude-transcripts.json. Their global cutoffs and locks remain separate, but their Claude state
    is shared and can suppress activity or race during writes. Append the suffix to the complete path or hash the complete internal-config path.

samuelpatro:
All three addressed in a521201:

- `claudeTranscriptStatePath` now appends the suffix to the full internal config path, so `work.cfg`, `work.ini` and `work` each get their own state file (`work.cfg-ai-claude-transcripts.json` etc.). The default becomes `~/.wakatime/wakatime-internal.cfg-ai-claude-transcripts.json`. The path test covers all three names.
- Expected paths in `claude_internal_test.go` are built with `filepath.Join`.
- The unix-permission cases (0500 dir in the save-errors test, 0000 file in the load-unreadable test) are skipped on Windows with the repo's usual `runtime.GOOS` guard. The other save-error cases (parent is a regular file, target is a directory) stay platform-independent and still run everywhere.

Verified with `GOOS=windows go test -c ./pkg/ai/` plus the usual `go test` and golangci-lint locally.

alanhamlett:
• Found three issues:

  - [P2] Fractional timestamps can still lose equal-timestamp records — claude.go:1501–1504
    The checkpoint timestamp passes through float64, which can round slightly past the original timestamp. On the next scan, a newly appended record with that same original timestamp fails the cutoff check before reaching
    count-based deduplication, discarding its heartbeat and usage. Preserve the original timestamp or normalize both sides consistently. The added whole-second examples miss this case.

  - [P2] Existing checkpoints prevent backfills after resetting the global cutoff — claude.go:463–468
    After one sync, moving ai_logs_last_parsed_at backward skips unchanged transcripts entirely. Changed transcripts also resume from their stored checkpoint instead of the requested earlier cutoff. Consequently, the reset-
    based backfill workflow described in the PR no longer works for previously tracked files. An explicit reset needs to invalidate or bypass the checkpoints.

  - [P2] Older Claude heartbeats can move the shared cutoff backward — claude.go:391
    This change allows Claude to return heartbeats older than the global cutoff, but WithAISync still unconditionally replaces that cutoff with the newest heartbeat returned during the current run. For example, a global
    cutoff of 13:00 becomes 12:30 if a slower Claude transcript supplies the only new heartbeat. Other parsers then reparse and resend activity already covered by the previous cutoff. Keep the shared watermark monotonic
    while advancing Claude’s checkpoints independently.

Issue #1365, opened 2026-04-20
Faulty user agent strings when doing manual and AI coding in parallel on different projects


muety:
I observed the following behavior: say you have some AI agent (Claude Code in my case) work on project A, while you, yourself, are working on (the entirely different and unrelated) project B in parallel in PyCharm (for example). Then heartbeats related to the AI coding (project A) will end up with a user agent string like this: 

```
wakatime/v2.5.0 (linux-6.19.12-200.fc43.x86_64-unknown) go1.25.9 Claude/2.1.114 pycharm/2026.1 pycharm-wakatime/16.0.0
```

It includes Claude as the AI agent plugin _and_ PyCharm as the IDE plugin, even though the PyCharm coding work was completely unrelated to the project that heartbeat refers to.

Without diving too deep into the code, I'd assume the cause for that issue lies somewhere in the "fallback" logic of [`aiUserAgent()`](https://github.com/wakatime/wakatime-cli/blob/233d35dce94f2e215ad6f7577ff36632c0cb9e4c/pkg/ai/ai.go#L411).

PR #1552, opened 2026-09-02
Scope app-heartbeat entity assignment to its own AI session

### LINKED ISSUE #1551: App heartbeats can inherit a file from another session, sending time to the wrong project
When several Claude Code sessions run concurrently in different projects, some AI heartbeats are recorded against a project that session never touched.

**What happens**

`replaceAppHeartbeats` (`pkg/ai/ai.go`) converts app-type heartbeats into file heartbeats by assigning them the last file entity seen while walking the slice. It is called on the combined output of `parseAIHeartbeats`, which concatenates every parser and every transcript into one flat slice — there is no per-session grouping and no sort. A session pulse from one transcript can therefore inherit a file path from a different transcript: it keeps its own timestamp and `AISession`, but points at a file in another project.

The same applies to the seed loop, which takes the first file entity in the whole slice for pulses that appear before any file heartbeat.

**Evidence**

From ~3,700 heartbeats, 21 sessions were flagged as touching more than one project. Checking each heartbeat's entity against its own session's transcript:

- 50 `(session, entity)` pairs where the file path appears **nowhere** in that session's own transcript
- all 50 of those paths **do** appear in a different transcript
- 37 heartbeats attributed to a project the session never worked in

Sessions that genuinely read across repositories look different — there the path is present in their own transcript, as expected.

**Impact**

Time and tokens land on the wrong project. It is easy to miss because each stray is only 1–3 heartbeats, but that is enough to make a single-project session look like it spanned several, and it silently moves time between unrelated projects.

**Related**

- #1365 was the same cross-project contamination for the `User-Agent` field, fixed in #1366 for that field only.
- #1548 / #1549 address concurrent-session state in this area, but do not touch this code path.

**Environment**:

- OS: linux
- Platform: amd64
- Version: v2.24.4; also confirmed unchanged on `develop` at 5fb9caa

**Logs**:

Nothing is logged — the heartbeats are sent successfully, just with the wrong entity.

imajus:
## Summary

- `replaceAppHeartbeats` rewrote app-type heartbeats (session pulses with no file entity) using a single rolling file entity walked across the entire parsed batch, which is a flat concatenation of every parser and every transcript. An app pulse from one session could inherit the trailing file entity of an unrelated session parsed right before it, sending its time and tokens to the wrong project.
- Entity is now tracked per `AISession`. A pulse only picks up a file from its own session; if that session has no file heartbeat anywhere in the batch, the pulse is left unconverted instead of borrowing someone else's.
- Same root cause as #1365 (fixed for `User-Agent` in #1366), but for the `Entity`/`EntityType` fields, which weren't touched by that fix.

Fixes #1551.

## Test plan

- [x] Added `TestRemoveNoopHeartbeatsDoesNotLeakEntityAcrossSessions`, verified it fails against the pre-fix code (wrong entity assigned, wrong heartbeat count) and passes against this change
- [x] Existing `pkg/ai` tests pass unchanged
- [x] `go test -race ./...` — all green except `TestSendHeartbeats_ExtraHeartbeatsNestedError`, which is a pre-existing race on `develop` unrelated to this change (confirmed by running it against `develop` before this patch)
- [x] `golangci-lint run ./...` — 0 issues


LOve-luST-LOST (commented):
@claude

alanhamlett (approved):

